PR #13, opened 2024-01-09
IR-68: Improve integration tests

ushkarev:
Previously, stubbing did not work for DPS’ frontend components api as the url was never set to point to wiremock.

Now, in most tests the fallback header and footer is used (to make signing in possible) but a new test exists to ensure html, css and javascript is correctly loaded.


feature.env
```diff
@@ -10,6 +10,7 @@ TOKEN_VERIFICATION_API_URL=http://localhost:9091/verification
 NOMIS_USER_ROLES_API_URL=http://localhost:9091/nomisUserRolesApi
 HMPPS_PRISON_API_URL=http://localhost:9091/prisonApi
 OFFENDER_SEARCH_API_URL=http://localhost:9091/offenderSearchApi
+COMPONENT_API_URL=http://localhost:9091/frontendComponents
```
ushkarev, line 13: ☝️ without this, frontend component stubbing had no effect

xoen (approved):

PR #130, opened 2024-08-06
IR-396: Incident Reporting client methods to get reports with details

xoen:
Added methods and types to get reports with details (e.g. staff/prisoners involvement, questions, etc...)

**NOTE**: the API has distinct "types" for Questions/Responses and historical ones - possibly because they're baked by different DB tables on the JPA side of things - but their format is the same so the UI types are the same.
This is simpler. We can always introduce more types when/if these will ever diverge.

Types for "enums" values like staff/prisoner roles, prisoner involvement outcome or correction request reasons are still defined as string (although we have a type alias) as they will likely change - we may also _possibly_/maybe move these into the UI from the API. But for now aliases to string will do.

### DIFF
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -79,8 +79,8 @@ export type ReportWithDetails = ReportBasic & {
   // TODO: questions
   // TODO: history
   historyOfStatuses: HistoricStatus[]
-  // TODO: staffInvolved
-  // TODO: prisonersInvolved
+  staffInvolved: StaffInvolvement[]
+  prisonersInvolved: PrisonerInvolvement[]
   // TODO: correctionRequests
 }
 
@@ -89,6 +89,12 @@ export type ReportType = string
 // TODO: Add enums?
 export type ReportStatus = string
 export type ReportSource = 'DPS' | 'NOMIS'
+// TODO: Add enums?
+export type StaffRole = string
+// TODO: Add enums?
+export type PrisonerRole = string
+// TODO: Add enums?
+export type PrisonerInvolvementOutcome = string
 
 export type GetEventsParams = {
   prisonId: string
@@ -118,6 +124,19 @@ export type HistoricStatus = {
   status: ReportStatus
   changedAt: Date
   changedBy: string
+}
+
+export type StaffInvolvement = {
+  staffUsername: string
+  staffRole: StaffRole
+  comment: string | null
+}
+
+export type PrisonerInvolvement = {
+  prisonerNumber: string
+  prisonerRole: PrisonerRole
+  outcome: PrisonerInvolvementOutcome | null
+  comment: string | null
 }
 
 export class IncidentReportingApi extends RestClient {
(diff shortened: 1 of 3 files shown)


server/data/incidentReportingApi.ts
```diff
@@ -122,6 +122,20 @@ export type PaginationSortingParams = {
...
 }
 
+export type Question = {
+  code: string
+  question: string
+  responses: Response[]
+  additionalInformation: string | null
+}
+
+export type Response = {
```
xoen, line 132: `Response` is a potentially bad/sneaky name because of other meanings/classes (e.g. `superagent` also has a `Response` type)

server/data/incidentReportingApi.ts
```diff
@@ -36,7 +40,7 @@ export type Paginated<T> = {
 export type PaginatedEventsWithBasicReports = Paginated<EventWithBasicReports>
 export type PaginatedBasicReports = Paginated<ReportBasic>
 
-export type EventWithBasicReports = {
+export type Event = {
```
xoen, line 43: `Event` is another potentially bad/ambiguous name (I was almost tempted to rename to `Incident` as this is really the parent incident).

The type `Event` [is already a thing in JS](https://developer.mozilla.org/en-US/docs/Web/API/Event) and there are also events in Node.js.

If that's the name in the domain so be it (although I still think there are really "incidents") but it's something to be wary.

server/routes/index.ts
```diff
@@ -30,6 +31,11 @@ export default function routes(service: Services): Router {
       // sort: ['eventDateAndTime,ASC'],
     })
 
+    // No authorisation at this point, no data shown in production
+    if (config.environment === 'prod') {
```
xoen, line 35: FYI: @jonesj2193 / @ushkarev I've added the simple logic we discussed about to _**not**_ show data in production. We'll do _something_ similar for the other pages.

server/data/incidentReportingApi.ts
```diff
@@ -229,4 +301,16 @@ export class IncidentReportingApi extends RestClient {
       path: `/incident-reports/reference/${reference}`,
     }).then(convertBasicReportDates)
   }
+
+  getReportWithDetailsById(id: string): Promise<ReportWithDetails> {
+    return this.get<DatesAsStrings<ReportWithDetails>>({
+      path: `/incident-reports/${id}/with-details`,
```
ushkarev, line 307: it's probs a good idea to escape `id` for url even if we expect to have control over the contents. i mean what if a user could go to a page `/some-exciting-url/report/[id]` and that id gets passed directly to the api… a good situation would be a boring 404, a worse situation a 500.

same point with `reference` below

server/routes/index.ts
ushkarev: cool. could throw a 404. doesn't really matter!

server/data/incidentReportingApi.ts
xoen: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/130/commits/622d47b92c3a72975a11465b59c531c864642244

PR #131, opened 2024-08-07
IR-375: Placeholder pages to display an incident/a report

xoen:
⚠️ **NOTE**: Not from designs (which are still WIP). Main purpose is to have a way to browse/check data without going straight to DB.

The pages may use further polishing but I think this is good enough to be merged before further tweaking.


server/views/pages/incident.njk
```diff
@@ -0,0 +1,108 @@
+{% extends "../partials/layout.njk" %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% set pageTitle = applicationName + " - Incident" %}
+{% set mainClasses = "app-container govuk-body" %}
```
ushkarev, line 4: not a big deal for throw-away pages, but `app-container` is not a valid class

server/views/pages/report.njk
```diff
@@ -0,0 +1,133 @@
...
+      }
+    ]
+  ]
+  }) }}
+
+  <h2>Prisoners involved</h2>
+  {%- if report.prisonersInvolved.length > 0 -%}
+    <ul>
+      {%- for prisonerInvolved in report.prisonersInvolved -%}
+        <li>{{ prisonerInvolved.prisonerNumber }} ({{ prisonerInvolved.prisonerRole }})</li>
```
ushkarev, line 76: should we include outcome? might help to spot whether it's been used in the past

ushkarev (dismissed):
some tiny comments, but this is fine for just viewing some data!

server/views/pages/incident.njk
xoen: 👌 [Removed non-existent CSS class](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/131/commits/6df6f1e61d51fee35df3c19d7ac48989c5dd289e)

ushkarev (dismissed):

server/views/pages/report.njk
xoen: Maybe, bear in mind these pages don't show everything-everything.

I've tweaked the page to show the prisoner outcomes as well, although I couldn't quickly think of any clever way of displaying this (but also it doesn't matter at this stage)

[Show prisoner involvement outcome](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/131/commits/88b71990fd36c82bd5eae5f7f1574541ecd6eb67)

ushkarev (approved):

PR #136, opened 2024-08-09
Added a page to view all incidents that can then be navigated to indi…

jonesj2193:
…vidual ones. Also added username lookup function to retrieve user's names from api using their system usernames

### DIFF
--- a/server/data/manageUsersApiClient.ts
+++ b/server/data/manageUsersApiClient.ts
@@ -22,4 +22,8 @@ export default class ManageUsersApiClient {
     logger.info('Getting user details: calling HMPPS Manage Users Api')
     return ManageUsersApiClient.restClient(token).get<User>({ path: '/users/me' })
   }
+
+  getNamedUser(token: string, username: string): Promise<User> {
+    return ManageUsersApiClient.restClient(token).get<User>({ path: `/users/${username}` })
+  }
 }
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -37,12 +37,7 @@ export default function routes(service: Services): Router {
     }
 
     const event = await incidentReportingApi.getEventById(id)
-
-    const inputUsernames = []
-    for (const report of event.reports) {
-      inputUsernames.push(report.reportedBy)
-    }
-
+    const inputUsernames = event.reports.map(report => report.modifiedBy)
     const nameMappings = await makeUsernameLookup(manageUsersApiClient, systemToken, inputUsernames)
 
     res.render('pages/incident', { event, nameMappings })
@@ -62,7 +57,7 @@ export default function routes(service: Services): Router {
       throw new NotFound()
     }
 
-    const report = await incidentReportingApi.getReportWithDetailsById(id
+    const report = await incidentReportingApi.getReportWithDetailsById(id)
     const reportedBy = (await manageUsersApiClient.getNamedUser(systemToken, report.reportedBy))?.name
     const prisonerNumbers = report.prisonersInvolved.map(pi => pi.prisonerNumber)
     const prisonersLookup = await offenderSearchApi.getPrisoners(prisonerNumbers)
@@ -85,11 +80,7 @@ export default function routes(service: Services): Router {
       })
     )?.content
 
-    const inputUsernames = []
-    for (const incident of incidents) {
-      inputUsernames.push(incident.modifiedBy)
-    }
-
+    const inputUsernames = incidents.map(incident => incident.modifiedBy)
     const nameMappings = await makeUsernameLookup(manageUsersApiClient, systemToken, inputUsernames)
 
     res.render('pages/showIncidents', { incidents, nameMappings })
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -8,6 +8,8 @@ import RedisTokenStore from '../data/tokenStore/redisTokenStore'
 import { createRedisClient } from '../data/redisClient'
 import { IncidentReportingApi } from '../data/incidentReportingApi'
 import config from '../config'
+import ManageUsersApiClient, { type User } from '../data/manageUsersApiClient'
+import { makeUsernameLookup } from '../utils/utils'
 
 const hmppsAuthClient = new HmppsAuthClient(new RedisTokenStore(createRedisClient()))
 
@@ -20,32 +22,13 @@ export default function routes(service: Services): Router {
     res.render('pages/index')
   })
 
-  get('/events', async (req, res, next) => {
-    const { user } = res.locals
-    const systemToken = await hmppsAuthClient.getSystemClientToken(user.username)
-    const incidentReportingApi = new IncidentReportingApi(systemToken)
-
-    const response = await incidentReportingApi.getEvents({
-      prisonId: user.activeCaseLoadId,
-      // eventDateFrom: new Date('2024-07-30'),
-      // eventDateUntil: new Date('2024-07-30'),
-      // sort: ['eventDateAndTime,ASC'],
-    })
-
-    // No authorisation at this point, no data shown in production
-    if (config.environment === 'prod') {
-      response.content = []
-    }
-
-    res.render('pages/events', { response })
-  })
-
   get('/incident/:id', async (req, res, next) => {
     const { id } = req.params
 
     const { user } = res.locals
     const systemToken = await hmppsAuthClient.getSystemClientToken(user.username)
     const incidentReportingApi = new IncidentReportingApi(systemToken)
+    const manageUsersApiClient = new ManageUsersApiClient()
 
     // No authorisation at this point, no data shown in production
     if (config.environment === 'prod') {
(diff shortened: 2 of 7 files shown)


server/routes/index.ts
```diff
@@ -8,6 +8,8 @@ import RedisTokenStore from '../data/tokenStore/redisTokenStore'
 import { createRedisClient } from '../data/redisClient'
 import { IncidentReportingApi } from '../data/incidentReportingApi'
 import config from '../config'
+import ManageUsersApiClient, { type User } from '../data/manageUsersApiClient'
```
xoen, line 11: I think the linter is complaining about the `User` not being used (so you can probably remove the import?)

server/data/manageUsersApiClient.ts
```diff
@@ -22,4 +22,8 @@ export default class ManageUsersApiClient {
     logger.info('Getting user details: calling HMPPS Manage Users Api')
     return ManageUsersApiClient.restClient(token).get<User>({ path: '/users/me' })
   }
+
+  getNamedUser(token: string, username: string): Promise<User> {
```
xoen, line 26: I wouldn't mind if this method was simply called `getUser()` :)

server/routes/index.ts
```diff
@@ -63,15 +53,42 @@ export default function routes(service: Services): Router {
...
+        // eventDateFrom: new Date('2024-07-30'),
+        // eventDateUntil: new Date('2024-07-30'),
+        // sort: ['eventDateAndTime,ASC'],
+      })
+    )?.content
+
+    const inputUsernames = []
+    for (const incident of incidents) {
+      inputUsernames.push(incident.modifiedBy)
+    }
```
xoen, line 87: This could be _something like_ (haven't tested but should be it): 

```JavaScript
const usernames = incidents.map(incident => incident.modifiedBy)
```

server/routes/index.ts
```diff
@@ -63,15 +53,42 @@ export default function routes(service: Services): Router {
...
+
+    const nameMappings = await makeUsernameLookup(manageUsersApiClient, systemToken, inputUsernames)
```
xoen, line 89: I do wonder if we could have a method in the `manageUsersApiClient` that gets the list of usernames and returns the mapping for you (but I'd map to the whole `User` to make the client more generic in case we need to use other information in other places)

Similar to what I've done for getting the prisoners names:

https://github.com/ministryofjustice/hmpps-incident-reporting/pull/135/files#diff-a5514f8aaeb918d8c1f5e7d5855b6179fcbb73ac46dcef3784aaea9e4c5c3d80R15-R28

That would make the Manage Users client more user-friendly and you don't need an utility function nor pass the client around as an argument :)

server/data/manageUsersApiClient.ts
jonesj2193: I named it `getNamedUser` as there is already a function in that class that is called `getUser` which gets the current user's profile using their token.

server/routes/index.ts
jonesj2193: removed it now. think I needed it at some point when I was putting stuff together

server/routes/index.ts
jonesj2193: replaced these now

server/routes/index.ts
jonesj2193: So do you mean to effectively move the function I created into the `manageUsersApiClient` and keep the whole `user` and not just the name? Just want to clarify as there currently isn't an API call to retrieve multiple users at once, so it would still need to be the same API call done for each username. 
Happy to look to implement this, just want to clarify first.

ushkarev: yeah exactly that! makes it more reusable

server/views/pages/showIncidents.njk
```diff
@@ -0,0 +1,47 @@
...
+{% set mainClasses = "app-container govuk-body" %}
+
+{% from "govuk/components/table/macro.njk" import govukTable %}
+
+{% block content %}
+  {% set rows = [] %}
+  {% for inc in incidents %}
+    {% set rows = (rows.push(
+      [
+      { html: "<a href='/incident/" + inc.id + "/'>" + inc.title + "</a>" },
```
ushkarev, line 12: when generating html, it's always important to escape variables – title might contain special characters. if we're being super-carefuly, id could be escaped but it's safe for now

you could do something like this:
```nunjucks
{% set title %}{{ title }}{% endset %}
```
and 
```javascript
{ html: `<a href="/incident/${inc.id}">${title}</a>` }
```

server/routes/index.ts
```diff
@@ -72,10 +58,32 @@ export default function routes(service: Services): Router {
...
+    const reportedBy = (await manageUsersApiClient.getNamedUser(systemToken, report.reportedBy))?.name
     const prisonerNumbers = report.prisonersInvolved.map(pi => pi.prisonerNumber)
     const prisonersLookup = await offenderSearchApi.getPrisoners(prisonerNumbers)
 
-    res.render('pages/report', { report, prisonersLookup })
+    res.render('pages/report', { report, prisonersLookup, reportedBy })
+  })
+
+  get('/incidents', async (req, res, next) => {
+    const { user } = res.locals
```
ushkarev, line 69: we ought to still prevent anyone from reaching this page in production with this from the deleted route

```javascript
    // No authorisation at this point, no data shown in production
    if (config.environment === 'prod') {
      response.content = []
    }
```

server/routes/index.ts
```diff
@@ -72,10 +58,32 @@ export default function routes(service: Services): Router {
...
+    const manageUsersApiClient = new ManageUsersApiClient()
+
+    const incidents = (
+      await incidentReportingApi.getEvents({
+        // prisonId: user.activeCaseLoadId,
+        // eventDateFrom: new Date('2024-07-30'),
+        // eventDateUntil: new Date('2024-07-30'),
+        // sort: ['eventDateAndTime,ASC'],
+      })
+    )?.content
```
xoen, line 81: I reckon the template should get the whole thing - not just content/the list of incidents - as they're needed to show pagination.

server/routes/index.ts
```diff
@@ -72,10 +58,32 @@ export default function routes(service: Services): Router {
     }
 
     const report = await incidentReportingApi.getReportWithDetailsById(id)
+    const reportedBy = (await manageUsersApiClient.getNamedUser(systemToken, report.reportedBy))?.name
```
xoen, line 61: Don't worry about the report page too much, but once this is merged I can also show the names of the staff involved.

server/routes/index.ts
```diff
@@ -21,41 +23,24 @@ export default function routes(service: Services): Router {
...
+    const manageUsersApiClient = new ManageUsersApiClient()
 
     // No authorisation at this point, no data shown in production
     if (config.environment === 'prod') {
       throw new NotFound()
     }
 
     const event = await incidentReportingApi.getEventById(id)
+    const inputUsernames = event.reports.map(report => report.modifiedBy)
+    const nameMappings = await makeUsernameLookup(manageUsersApiClient, systemToken, inputUsernames)
```
xoen, line 41: We could call this `staffLookup` for consistency with the variable for the `prisonersLookup` :)

server/views/pages/incident.njk
```diff
@@ -83,7 +83,7 @@
             text: "Reported by"
           },
           {
-            text: report.reportedBy
+            text: nameMappings[report.reportedBy]
```
xoen, line 86: (Nunjacks has the annoying tendency to fail silently)

It's probably a good idea to fallback in case for whatever reason the username was not found.

I know we had in the past shown "System" (or something like that) for system usernames that may not be real usernames.

But I think for now you could keep this simple and just display the original username, something like this could work:

```
text: nameMappings[report.reportedBy] | default(report.reportedBy)
```

although I imagine a plain `or` would also do:

```
text: nameMappings[report.reportedBy] or report.reportedBy
```

server/routes/index.ts
jonesj2193: made some adjustments to keep the whole response

server/views/pages/showIncidents.njk
jonesj2193: think I have this in place now and seems to work...

server/routes/index.ts
jonesj2193: I have done this. It is now as `getUsers` in the `userService`.

server/data/manageUsersApiClient.ts
jonesj2193: This is now resolved in the moving of the function.

server/routes/index.ts
jonesj2193: this is not needed anymore due to the new `userService.getUsers` function

server/routes/index.ts
xoen: Addressed in line 70.

server/utils/utils.ts
```diff
@@ -1,3 +1,5 @@
+import ManageUsersApiClient from '../data/manageUsersApiClient'
```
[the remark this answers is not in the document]
ushkarev: otherwise looks fine!

server/routes/index.ts
xoen: 🤔 I think eventually we'd need pagination to work as when we'll have many incidents this will only show the first page (and there will be no way to show the subsequent pages).

But maybe this could go in without that for now? And maybe added separately - @ushkarev what do you reckon?

server/utils/utils.ts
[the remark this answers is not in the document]
jonesj2193: On it. Just pushed removing it

server/routes/index.ts
ushkarev: it doesn't have to be done in this pr, but i think some basic filtering and pagination are worth adding. since we would need to deal with that somehow in future anyway and there's nothing more concrete to build 🤷

PR #138, opened 2024-08-13
IR-374: Add pagination to incident list debug view

ushkarev:
… so that the components are already in place for future work.


server/views/pages/debug/reportDetails.njk
```diff
@@ -113,7 +129,22 @@
     {%- set questionsRows = [] -%}
     {%- for question in report.questions -%}
       {%- if question.responses.length > 0 -%}
-        {%- set responses = question.responses | join('<br />', "response") -%}
+        {% set responses %}
```
xoen, line 132: 👍

server/views/pages/debug/reportDetails.njk
```diff
@@ -138,9 +169,11 @@
   <h2>Correction requests</h2>
   {%- if report.correctionRequests.length > 0 -%}
     <ul>
-      {%- for correctionRequests in report.correctionRequests -%}
+      {%- for correctionRequest in report.correctionRequests -%}
```
xoen, line 172: oops, an `s` slipped :)

server/views/pages/debug/reportDetails.njk
ushkarev: this is cos the date & comment might be worth showing

PR #139, opened 2024-08-13
IR-404: Test API clients

ushkarev:
We only have debugging routes/views at present and little of their code is likely to remain in future.

The incident-reporting API client, however, will still be around so ought to have some tests!
As part of this, there is some test event and report generation being added – these will prove necessary for mocking out the API.

### DIFF
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -139,9 +139,10 @@ export type Question = {
 
 export type Response = {
   response: string
+  // recordedDate: Date | null // TODO: DatesAsStrings does not work on optional dates
+  additionalInformation: string | null
   recordedBy: string
   recordedAt: Date
-  additionalInformation: string | null
 }
 
 export type HistoricReport = {
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -139,7 +139,7 @@ export type Question = {
 
 export type Response = {
   response: string
-  // recordedDate: Date | null // TODO: DatesAsStrings does not work on optional dates
+  responseDate: Date | null
   additionalInformation: string | null
   recordedBy: string
   recordedAt: Date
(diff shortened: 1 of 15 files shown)


server/utils/format.ts
```diff
@@ -38,4 +53,13 @@ export default {
       timeZone: 'Europe/London',
     })
   },
+
+  /**
+   * Formats dates in ISO style, used when calling APIs
+   *
+   * Example: `2024-07-30`
+   */
+  isoDate(date: Date): string {
```
xoen, line 62: `isoDate()` is a better name :)

I think this is fine, but I guess the `&&` may silently "fail" and return something invalid, e.g. `false && ` would return `false` rather than an empty string 🤔 (before it would throw an exception)

We could leave as is :) - or add a check for at least `undefined`/`null` - but I don't think we have to be super-rigorous on this one, JS is a bit of a mess when it comes to type checking (anything is an `object`)

server/data/incidentReportingApi.ts
```diff
@@ -28,7 +28,9 @@ export class ErrorResponse {
   }
 }
 
-export type Paginated<T> = {
+export const defaultPageSize = 20
+
+export interface Page<T> {
```
xoen, line 44: I suppose it is a single page...

server/utils/utils.ts
```diff
@@ -52,6 +52,34 @@ export const initialiseName = (fullName?: string): string | null => {
...
+  const d = parseInt(day, 10)
+  if (Number.isSafeInteger(y) && m >= 1 && m <= 12 && d >= 1 && d <= 31) {
+    const date = new Date(y, m - 1, d)
+    if (date) return date
+  }
+  throw new Error('Invalid date')
+}
+
+/** Find field error in error summary list */
+export const findFieldInErrorSummary = (list: ErrorSummaryItem[], formFieldId: string): { text: string } | null => {
```
xoen, line 71: I'm surprised these 2 functions show in the GH diffs as they're now in `main` 🤔

xoen (approved):
Nice one 👍

server/utils/utils.ts
ushkarev: i think it shows in the diff cos there was a merge conflict…?

PR #141, opened 2024-08-15
IR=376: Allow filtering of incidents by dates

xoen:
**NOTE**: This is still undesigned, these views are mainly for debug purposes.

Using HMPPS Date picker component (as used in Incentives services)


server/utils/format.ts
```diff
@@ -38,4 +38,14 @@ export default {
       timeZone: 'Europe/London',
     })
   },
+
+  /** Formats a date, e.g. 31/10/2021 */
+  formDate(date: Date): string {
```
ushkarev, line 43: isn't this the same as `shortDate`?

xoen: Looks like it actually, aside from the check for null/undefined (which I think is useful).

Me thinks I'll take combine the too, retain that check and replace everything with one, but I'll have to double check things tomorrow 👍

xoen: Or just use the existing function 😋 🤦‍♂️

xoen: [Removed unnecessary utility function](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/141/commits/a6fd4b4960b3d99072ce56a20f9d36da1146e6d8)

ushkarev (dismissed):

ushkarev (approved):

PR #143, opened 2024-08-19
IR-376: Show prison names

xoen:
Show prison names instead of their IDs.

Also refactored things a little to move some of the common set-up logic related to APIs in a couple of small middleware.

### DIFF
--- a/integration_tests/tsconfig.json
+++ b/integration_tests/tsconfig.json
@@ -10,8 +10,8 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "sourceMap": false,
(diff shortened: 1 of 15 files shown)


integration_tests/tsconfig.json
```diff
@@ -10,8 +10,8 @@
     "esModuleInterop": true,
     "skipLibCheck": true,
     "sourceMap": false,
-    "typeRoots": ["../server/@types", "../node_modules/@types", "../node_modules"],
+    "typeRoots": ["../node_modules/@types", "../node_modules", "../server/@types"]
```
xoen, line 13: Order seems to matter.

ushkarev (approved):

PR #144, opened 2024-08-20
IR-411: Retrieve questionnaires from prison-api

ushkarev:
Consumes [newly exposed NOMIS api](https://github.com/ministryofjustice/prison-api/pull/2260) to dump CSV files with incident report configuration.


server/data/offenderSearchApi.test.ts
```diff
@@ -21,15 +21,15 @@ describe('offenderSearchApi', () => {
...
+    it('should de-duplicate input prisoner numbers', async () => {
       fakeApiClient.post('/prisoner-search/prisoner-numbers').reply(200, (_uri, requestBody) => {
         const request = requestBody as { prisonerNumbers: string[] }
         expect(request.prisonerNumbers).toHaveLength(3)
         return [chris, barry, andrew]
       })
 
       const responseFuture = apiClient.getPrisoners(['A1111AA', 'A2222BB', 'A1111AA', 'A3333CC'])
-      expect(responseFuture).resolves.toEqual({
+      await expect(responseFuture).resolves.toEqual({
```
xoen, line 32: To be honest we could drop the `resolves` (unless I'm missing something), e.g.

```JS
const response = await apiClient.getPrisoners(['A1111AA', 'A2222BB', 'A1111AA', 'A3333CC'])
await expect(responseFuture).resolves.toEqual({...})
```

No need to change as it's minor tiny thing.

ushkarev: fair point. will have a look as part of next pr; might make things tidier

server/data/prisonApi.ts
```diff
@@ -8,6 +8,78 @@ export type Prison = {
...
+  /** Date question was retired */
+  questionExpiryDate?: Date
+  /** Indicate multiple responses can be given */
+  multipleAnswerFlag: boolean
+  /** Order list of answers */
+  answers: Answer[]
+}
+
+/** Incident Type Answer */
+interface Answer {
```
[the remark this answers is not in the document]
ushkarev: fair point. i didn't love the names in prison-api, but maybe i can find a compromise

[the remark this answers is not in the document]
ushkarev: in fact, i've used exactly your suggestion 51b6f4aab5412d75236b80b19f8287efb85eeece

server/data/prisonApi.ts
```diff
@@ -133,4 +146,39 @@ export class PrisonApi extends RestClient {
...
+      return {
+        ...referenceCode,
+        expiredDate: referenceCode.expiredDate && new Date(referenceCode.expiredDate),
+        subCodes: (referenceCode.subCodes ?? []).map(parseDates),
+      }
+    }
+
+    return this.get<DatesAsStrings<ReferenceCode[]>>({
+      path: `/api/reference-domains/domains/${encodeURIComponent(domain)}/codes`,
+    }).then(codes => codes.map(parseDates).sort((code1, code2) => (code1.listSeq ?? 0) - (code2.listSeq ?? 0)))
```
xoen: isn't `listSeq` always there 🤔?

ushkarev: no. lol

xoen: 😬 - in that case should the `ReferenceCode` interface reflect that? e.g. `listSeq?: number`

server/routes/debug.ts
```diff
@@ -89,14 +90,13 @@ export default function makeDebugRoutes(services: Services): Record<string, Requ
...
     async incidentDetails(req, res) {
-      const { userService } = services
-      const { incidentReportingApi, prisonApi } = res.locals.apis
-
       const { id } = req.params
       if (!id) {
         throw new NotFound()
       }
 
+      const { incidentReportingApi, prisonApi } = res.locals.apis
```
xoen: Actually, I've purposefully put this up at the beginning of the function as this are kinda the dependencies this function depends on, rather then scattering these around the function.

server/data/incidentReportingApi.ts
```diff
@@ -196,7 +196,7 @@ export class IncidentReportingApi extends RestClient {
     super('HMPPS Incident Reporting API', config.apis.hmppsIncidentReportingApi, systemToken)
   }
 
-  getEvents(
+  async getEvents(
```
xoen, line 199: 💅

server/routes/downloadNomisConfig.ts
```diff
@@ -2,22 +2,59 @@ import type { RequestHandler, Response } from 'express'
 
 import type { ReferenceCode } from '../data/prisonApi'
 
-function streamReferenceCodeResponse(fileName: string, referenceCodes: ReferenceCode[], res: Response): void {
+type CsvCellValue = string | number | boolean | null
+
+function streamCsvDownload(res: Response, fileName: string, lines: Generator<CsvCellValue[]>): void {
```
xoen, line 9: tbh, the generator could be an internal implementation of `streamCsvDownload()`, e.g.

```TS
// as before
streamCsvDownload(res, 'incident-types.csv', referenceCodes)
// [...]
for (const line of mapCodes(lines)) {
  // [...]
}
```

it could even be a nested function if we feel particularly courageous (still figuring out if they're a good idea or not 😝 )

ushkarev: i thought it made the row mapping function _bodies_ easier to read with separate yields for each row. could just as easily build a giant array of all rows. it's what i did at first but this looks a bit better i think.

server/routes/index.ts
```diff
@@ -27,6 +27,10 @@ export default function routes(services: Services): Router {
   const downloadNomisConfigRoutes = makeDownloadNomisConfigRoutes()
   get('/nomis-report-config/incident-types.csv', downloadNomisConfigRoutes.incidentTypes)
   get('/nomis-report-config/incident-type/:type/questions.csv', downloadNomisConfigRoutes.incidentTypeQuestions)
+  get(
+    '/nomis-report-config/incident-type/:type/prisoner-roles.csv',
```
xoen, line 31: But prisoner roles are not specific of any Report type tho no?

ushkarev: it appears that they _are_. there's a global set of roles that exist, but what's available per-question is given here.
incident type config

xoen: Ah OK, I didn't realise this 👌

xoen (approved):

server/routes/index.ts
ushkarev: me neither

PR #160, opened 2024-09-04
Adding Form wizard to repo

jonesj2193:
This includes using the form wizard to create a user transaction to create a new incident and another to edit an existing incident.


cypress.config.ts
```diff
@@ -38,6 +38,6 @@ export default defineConfig({
     baseUrl: 'http://localhost:3007',
     excludeSpecPattern: '**/!(*.cy).ts',
     specPattern: 'integration_tests/e2e/**/*.cy.{js,jsx,ts,tsx}',
-    supportFile: 'integration_tests/support/index.ts',
+    supportFile: 'integration_tests/support/createIncident.ts',
```
ushkarev, line 41: you don’t need to change this line; it’s causing integration tests to fail and we’re not yet making any new ones

server/utils/utils.ts
```diff
@@ -126,3 +128,9 @@ export const govukSelectSetSelected = (items: GovukSelectItem[], value: string):
     selected: 'value' in entry ? entry.value === value : entry.text === value,
   }))
 }
+
+const template = '{% from "govuk/components/input/macro.njk" import govukInput %}{{ govukInput(params) }}'
+
+export default function freeTextInput(params: object) {
+  return nunjucks.renderString(template, { params })
+}
```
ushkarev, line 136: what’s this utility for?

server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -0,0 +1,283 @@
+// Copied from https://github.com/ministryofjustice/hmpps-strengths-based-needs-assessments-ui/blob/c1d2e0ac64ce2be0ea0a27919cd34f8882b2d5c0/server/%40types/hmpo-form-wizard/index.d.ts
+
+/* eslint-disable max-classes-per-file */
+declare module 'hmpo-form-wizard' {
+  import Express from 'express'
+
+  // These enums have to live here because of TS/Jest and Enums work..  ¯\_(ツ)_/¯
+  // Also this ESLint override because of how TS/Eslint works.
+  // eslint-disable-next-line no-shadow
```
ushkarev, line 9: eslint is saying that you don’t need any of the `no-shadow` exemptions, but elsewhere you can just ignore the problem with `any` with
```javascript
// eslint-disable-next-line @typescript-eslint/no-explicit-any
```

cypress.config.ts
jonesj2193: I don't remember changing this, not sure if it somehow changes when I refactored something.
Thanks as I was going to ask how to fix this

server/utils/utils.ts
jonesj2193: So this is used to help add the conditional input entries for radio buttons. I was playing with it to get it working, but I didn't quite get there as it wasn't showing. So I left it for now to get other things done, but I will come back to this and likely use this function when I return to the other page designs.
I can remove it for now if you would like and can then re-add it when I go to add those conditional fields.

server/@types/hmpo-form-wizard/index.d.ts
jonesj2193: ah, I was trying:
`// eslint-disable-next-line no-explicit-any`
and it wasn't working. Will try the above thank you.

server/routes/index.ts
```diff
@@ -35,5 +37,9 @@ export default function routes(services: Services): Router {
   get('/nomis-report-config/prisoner-involvement-roles.csv', downloadNomisConfigRoutes.prisonerInvolvementRoles)
   get('/nomis-report-config/prisoner-involvement-outcome.csv', downloadNomisConfigRoutes.prisonerInvolvementOutcome)
 
+  router.use('/create-incident', createIncidentRouter)
+  router.use('/change-incident/:id/', changeIncidentRouter)
+  router.use('/report/:incidentId/add-prisoner/:prisonerId', changeIncidentRouter)
```
ushkarev, line 42: i feel that this needs the same `config.environment !== 'prod'` guard as other routes. what do you think?

jonesj2193: good point, I will move these now

server/utils/utils.ts
ushkarev: your call, i guess! i can't tell quite how it'll work 🤷

server/controllers/changeIncident/changeIncident.ts
```diff
@@ -0,0 +1,123 @@
...
+      const { user } = res.locals
+      const { incidentDate } = req.form.values
+      const { incidentTime } = req.form.values
+      const { incidentPrisonId } = req.form.values
+      const { incidentTitle } = req.form.values
+      const { incidentDescription } = req.form.values
+
+      const hmppsAuthClient = new HmppsAuthClient(new RedisTokenStore(createRedisClient()))
+      const systemToken = await hmppsAuthClient.getSystemClientToken(user.username)
+      const incidentReportingApi = new IncidentReportingApi(systemToken)
```
xoen, line 87: Not massively important for the sake of this PR but we added some middleware to reduce duplication and now you can easily get an instance of `IncidenReportingApi` (and others) from `res.locals.api` :)

e.g. https://github.com/ministryofjustice/hmpps-incident-reporting/blob/d39b44d9b85a82a438f5bcc35ca0712b95824212/server/routes/debug.ts#L112

Which turns this into a one line instead of 3 :)

(you can also find the system token in `res.locals.systemToken` but I don't think you'd need in this case if you get the `IncidentReportingApi` instance from `res.locals.apis`, e.g. https://github.com/ministryofjustice/hmpps-incident-reporting/blob/d39b44d9b85a82a438f5bcc35ca0712b95824212/server/routes/debug.ts#L121)

xoen: Incidentally the `const { user } = res.locals` line could also go if you get `res.locals.apis.incidentReportingApi` because you don't need to generate a system token explicitly and therefore you don't need the `user` variable.

server/data/incidentReportingApi.ts
```diff
@@ -193,6 +193,31 @@ export type CorrectionRequest = {
...
+  title: string | string[]
+  description: string | string[]
+  createNewEvent: boolean
+}
+
+export type UpdateIncident = {
+  incidentDateAndTime: string | string[]
+  prisonId: string | string[]
+  title: string | string[]
+  description: string | string[]
```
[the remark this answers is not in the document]
jonesj2193: so they are strings, and I set them to just be strings initially, but they come back as errors when trying to define them as the values that come back from the form are of type `string | string[]`. So that's why I have set them to be it. For now at least. But am open to a solution of how to fix that.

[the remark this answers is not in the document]
jonesj2193: I've sorted it just by explicitly forcing them to be strings on allocation

server/@types/express/index.d.ts
```diff
@@ -24,6 +26,16 @@ export declare global {
...
+      services?: Services
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      flash(): { [key: string]: any[] }
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      flash(message: string): any[]
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      flash(type: string, message: any[] | any): number
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      flash(type: string, format: string, ...args: any[]): number
+      canAccess: (permission: string) => boolean
```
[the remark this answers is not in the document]
ushkarev: i think it's quite misleading to leave it in. seeing as this is the type definition or the `req` object, both _we_ and typescript will assume that, for example, services can be accessed from it. but they cannot.

personally, i would remove it. but if you want it as a reminder, i'd urge including TODO comments to say that `flash` middleware is not installed, `canAccess` has not been written and `services` are not attached to this object.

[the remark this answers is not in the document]
ushkarev: furthermore, the absence of `canAccess` permeates to `protectRoute` – one wouldn't know that it's unusable given that the code of `protectRoute` makes sense and even that it's tests pass!

server/controllers/createIncident/createIncident.ts
```diff
@@ -0,0 +1,81 @@
...
+      backLink,
+      cancelLink: backLink,
+    }
+  }
+
+  async saveValues(req: FormWizard.Request, res: Response, next: NextFunction) {
+    try {
+      const { incidentType } = req.form.values
+      const { incidentDate } = req.form.values
+      const { incidentTime } = req.form.values
```
xoen, line 43: not a biggie, but you could probably destructure all of these at once, e.g.

```JavaScript
const { incidentType, incidentDate, incidentTime, incidentTitle, incidentPrisonId } = req.form.values
```

server/data/incidentReportingApi.ts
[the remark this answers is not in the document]
ushkarev: cool beans! it's absolutely right to force the types on the _other_ side because these endpoints definitely do not accept arrays.

server/controllers/createIncident/createIncident.ts
jonesj2193: good point, changed these now

server/@types/express/index.d.ts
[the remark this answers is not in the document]
jonesj2193: removed them and added a TODO.

ushkarev (approved):

PR #163, opened 2024-09-08
IR-439: Update incident-reporting API endpoints

ushkarev:
* the create and update report endpoints
    * return reports, not their respective inputs
    * need to convert dates in their responses
    * can use stronger date typing on their inputs
* new methods to list, add, update and delete:
    * staff involvement
    * prisoner involvement
    * correction requests
* new methods to list, add and delete questions


Mjwillis (approved):
LGTM

server/data/incidentReportingApi.ts
```diff
@@ -423,20 +423,38 @@ export class IncidentReportingApi extends RestClient {
...
 }
 
-class RelatedObjects<ResponseType> {
+type AddStaffInvolvement = {
+  staffUsername: string
+  staffRole: StaffRole
+  comment?: string
+}
+
+type AddPrisonerInvolvement = {
```
xoen: Should these also have the `*Request` suffix?

ushkarev: yes good idea. will add to all type. (there are more in subsequent commits)

ushkarev: blindly adding a `Request` suffixes has resulted in “AddQuestionWithResponsesRequest” … while it follows the pattern, does it sound too stupid?

server/data/incidentReportingApi.ts
```diff
@@ -473,6 +473,13 @@ export class IncidentReportingApi extends RestClient {
     })
     return questions.map(convertQuestionDates)
   }
+
+  async deleteLastQuestionWithResponses(reportId: string): Promise<Question[]> {
```
xoen, line 485: perhaps a slightly clearer name for this method could be `deleteLastQuestionAndItsResponses()`...more accurate description of what's doing, may help people not-so familiar with the API.

...but maybe pedantic?

server/data/incidentReportingApi.ts
xoen: a bit yeah 😅 - especially as there is a mixture or domain concept and HTTP concept in there 😬

xoen (approved):
nice work :)

PR #166, opened 2024-09-09
IR-68: Sync changes from upstream template

ushkarev:
Taken from:
- [hmpps-template-typescript#413](https://github.com/ministryofjustice/hmpps-template-typescript/pull/413)
- [hmpps-template-typescript#415](https://github.com/ministryofjustice/hmpps-template-typescript/pull/415) – except we load user details from an api not the token (as before) because we need more than what's available in it


xoen (approved):

ushkarev:
i'm unsure about incorporating [hmpps-template-typescript#412](https://github.com/ministryofjustice/hmpps-template-typescript/pull/412) because we would have to deal with a bunch of k8s secrets renaming and the auto-rotation process may be tricky

PR #167, opened 2024-09-09
IR-407: Tweaks to form wizard proof of concept

ushkarev:


server/middleware/protectRoute.ts
```diff
@@ -1,14 +1,18 @@
-import { RequestHandler } from 'express'
+import type { RequestHandler } from 'express'
+import { Forbidden } from 'http-errors'
 
-export default function protectRoute(permission: string): RequestHandler {
-  return async (req, res, next) => {
-    if (req.canAccess(permission)) {
+// TODO: why use this in place of the standard authorisationMiddleware?
```
xoen, line 4: That's a very good point, differences in implementations aside this is very similar to TS template's `authorisationMiddleware`...should we drop this one?

xoen: Especially when it comes to security/permissions I'd prefer to have a single, well-established way of doing something rather than maintain 2 versions of the "same" thing :)

xoen (approved):

PR #176, opened 2024-09-12
IR-497: Add a utility to check NOMIS report config validity

ushkarev:
This can be used as part of the import process of NOMIS configuration into our code base.

The key thing is to ensure that all routes through a report’s set of questions are valid: there are no loops.

Once we start making our own config for future report types, we can convert this to check _our_ config instead of NOMIS’.

### DIFF
--- a/server/reportConfiguration/analyseNomisReportConfiguration.ts
+++ b/server/reportConfiguration/analyseNomisReportConfiguration.ts
@@ -30,14 +30,31 @@ export function analyseNomisReportConfiguration(incidentTypes: readonly Incident
     types.add(incidentType.incidentType)
     questionCount += incidentType.questions.length
 
+    let firstQuestionId: number | null = null
+    const questionConnections = new Map<number, Set<number>>()
+
     incidentType.questions.forEach(question => {
       questionIds.add(question.questionnaireQueId)
       answerCount += question.answers.length
+
+      if (firstQuestionId === null) {
+        firstQuestionId = question.questionnaireQueId
+      }
+      questionConnections.set(
+        question.questionnaireQueId,
+        new Set(question.answers.map(answer => answer.nextQuestionnaireQueId)),
+      )
 
       question.answers.forEach(answer => {
         answerIds.add(answer.questionnaireAnsId)
       })
     })
+
+    try {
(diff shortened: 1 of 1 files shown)


xoen:
FYI: @ushkarev I've been working on doing validation on the DPS config and I'm cooking a PR which should be ready for review soon: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/184

I think some of the logic in there could be reused (I've made the data structure generic for this reason). I'm happy to 1) show you what I've done and 2) chat when you're back as I have ideas :)

server/reportConfiguration/analyseNomisReportConfiguration.ts
```diff
@@ -55,3 +72,37 @@ export function analyseNomisReportConfiguration(incidentTypes: readonly Incident
...
+    }
+    const nextQuestionIds = questionConnections.get(nextQuestionId)
+    if (!nextQuestionIds) {
+      throw new Error(`Question id ${nextQuestionId} is not within report`)
+    }
+    if (path.includes(nextQuestionId)) {
+      throw new Error(`Question cycle detected: ${path} loops back to ${nextQuestionId}`)
+    }
+    path.push(nextQuestionId)
+    const clonedPath = structuredClone(path)
```
xoen, line 89: I can't remember if I mentioned this to you....but I ran into problems with `structuredClone()` early on with one of the bigger report type config.

Surprisingly it was using a lot of memory, not sure if it was a bug in my logic - possible - but it wasn't happening when traversing other report configs. Something to bear in mind.

Also, I think you could use [the `Graph` data structure](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/main/server/data/incidentTypeConfiguration/directedGraphs.ts) here to check for cycles and other problems like I do when I validate the DPS config.

You'd need to do [something similar to this](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/main/server/data/incidentTypeConfiguration/validation.ts#L53-L65) but to generate the graph from the NOMIS config, then call `dfs()` to do the traversal and you can use the result to determine if there are cycles or unreachable nodes, etc...

ushkarev:
this PR was the predecessor for **DPS** config validation and that exists now. there's probably nothing here to keep

PR #178, opened 2024-09-16
IR-499: Questionnaire config

xoen:
- initial attempt at a format to encapsulate a given Report type questions/answers
- logic to convert NOMIS config for a type into this new format
- logic to write this new format into a TypeScript file
  - the config in these files have a type which is checked against the new format
- script to create/update the new config from the NOMIS data (extracted using `/download-report-config/nomis/incident-types.json` URL)
  - this script can be re-run when something in NOMIS changes or if/when the new DPS config format changes
- use newly generated `Type`/`NomisType` types (the type of a Report, e.g. `ASSAULT`/`ASSAULTS3`) instead of strings, so that TS can automatically check those
- generated config for production report types (under `/server/reportConfiguration/types`)
- refactored `listDownloadLinks`/`importDpsConstants` scripts


Mjwillis (approved):
LGTM

PR #179, opened 2024-09-16
IR-499: Added code to questions/answers

xoen:
As we tweak some of the questions/answers (for example to avoid SHOUT CASE while still retaining the original meaning) it's important to retain the original question/answer and use that internally.

This is because downstream services rely on these "codes" (e.g. `WHAT WAS THE METHOD OF CONCEALMENT`) to calculate stats, etc...even innocuous-tweaks can be breaking changes.

This is just adding a `code` field, which is currently populated from NOMIS config and matches the "label".

I'm adding this field now to reduce the risk of future confusion.

TL;DR: Questions/Answers "code" should never change, while the labels can be tweaked as long as the semantic/meaning doesn't change (e.g. we could retain the code for that question while changing the user-facing label to something like `What was the method of concealment?`)


Mjwillis (approved):
Lgtm - assume will convert to sentence case in due cause

PR #181, opened 2024-09-18
Fixing issue where error is thrown when creating or editing an incident

jonesj2193:
An issue has occurred where an error is thrown when trying to access the forms to create or edit an incident. 

The error is related to a missing cookie when checking the session as part of the form wizard. A previous PR removed the cookie parser from the package and app setup. Putting these back in appears to fix the problem.


jnioche-jd (approved):

ushkarev:
@jonesj2193 @jnioche-jd are you guys sure this is correct? the user’s session is not stored in a cookie. it might be that we need a tweak to the form wizard config to use the existing session correctly instead?

for one thing, having two different sessions for storing user data is confusing – what properties ends up in which? for another, using cookies for sessions *may* pose a bit of a security/privacy problem – is this new cookie session encrypted or at the very least signed for authenticity?

PR #184, opened 2024-09-23
IR-505: Config validation and visualisation

xoen:
- `updateNomisIncidentTypeConfigurations` script now checks the updated Incident Report Type config files for validation problems
  - we'll be able to see at a glance which report types are invalid and what are the issues
  - check starting question is not invalid (e.g. `null`) or unknown (e.g. a question ID for a question not present in questions set) or inactive
  - check for unknown questions, e.g. some answers lead to these question IDs but they're not in the given questions set
  - check for unreachable questions, e.g. no sequence of answers will ever be able to reach that question
  - check for cycles, e.g. somehow some answers lead to a question which was already answered
- when importing report types config SVG files with a the visual flow of the questions/answers are generates
  - example: [`FINDS.svg`](https://raw.githubusercontent.com/ministryofjustice/hmpps-incident-reporting/2c5c8af0cfee8d33f417f55aecf759c1412b6d84/server/reportConfiguration/types/FINDS.svg)
  - these files make it very easy to check for issues
  - it's much easier to wrap your head around a questionnaire flow by looking at the flowchart
  - non-technical members of the team should also be able to understand these
  - visualisation highlights inactive questions/answers
- changes to config format
  - added `active` flag for questions/answers
- more robust conversion of question ids to string (previously it could have raised an error if a question id was `undefined`)
- various refactoring
  - separation of concern between conversion to typescript and persistance/formatting of the file
- it should be possible to reuse the `Graph<T>` data structure to perform similar validation for the old NOMIS config
  - the data structure is generic over the question Id type for this reason (e.g. DPS question IDs are strings while NOMIS question IDs are numbers)

### How to review
This PR has a lot of changes and the commits iterated a lot. Review commit-by-commit is not advised.

The order of diff shown in GitHub is also probably the most natural.

I suggest you look at the whole diff but look at these files in the following top-down order:
- `scripts/updateNomisIncidentTypeConfigurations.ts` - start from the script and drill down
- `validateConfig()` in `../server/data/incidentTypeConfiguration/validation.ts`
- then move to `server/data/incidentTypeConfiguration/directedGraphs.ts` for the Directed Graph data structure used to check for unreachable/unknown questions and for cycles
- look at `toGraphviz()` in `../server/data/incidentTypeConfiguration/conversion.ts` for the logic to generate the questionnaire graphs (this data will then be converted to SVG in `persistance.ts` using graphviz's `dot` cli)
- there are a lot of files under `server/reportConfiguration/types`, these are autogenerated
- once you have a understanding of how things fit together it should be easier to look at the whole diff


Mjwillis (approved):
LGTM

server/utils/setUtils.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Returns a new set with the elements of A which are not in B
+ *
+ * @param a first Set
+ * @param b second Set
+ * @returns the Set with the difference A - B
+ */
+export function setDifference<T>(a: Set<T>, b: Set<T>): Set<T> {
```
ushkarev, line 8: might be worth noting that node 22 will have this inbuilt

PR #188, opened 2024-09-25
IR-506: Add `prisonerRoles` field to DPS incident report type config

xoen:
- re-generated `prisonerInvolvementRoles` constant to have `nomisCode`
- generated `NomisPrisonerInvolvementRole` type with valid NOMIS prisoner involvement role codes
- added `prisonerRoles` field to DPS incident report type config with allowed prisoner roles for that type
- when updating the incident report type config files also populate `prisonerRoles` field


server/routes/downloadReportConfig/nomis.test.ts
```diff
@@ -211,7 +211,7 @@ Question ID,Question sequence,Question list sequence,Question,Allows multiple an
               expiryDate: new Date(2022, 7, 20),
             },
             {
-              prisonerRole: 'LEAD',
```
xoen, line 214: I'm guessing this may have stand for "leader"? But I don't see this code in [the generated prisoner roles](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/188/commits/58e941489246144443fa7d4cf5d12dc1be718c02#diff-7cb234be6aab76e9d83dfaa2b0b7bb2c1dfb94e88661f8e227d1cbd0bd2e7bec).

ushkarev: i must've buggered this one up. no idea where i got it from.

server/data/incidentTypeConfiguration/types.ts
```diff
@@ -38,3 +40,13 @@ export interface AnswerConfiguration {
...
 }
+
+/** Report type prisoner role configuration */
+interface PrisonerRoleConfiguration {
+  /** Role type for this report type */
+  prisonerRole: PrisonerInvolvementRole
+  /** If only one prisoner can have this role in a given report */
+  onlyOneAllowed: boolean
+  /** Indicates this role is active */
+  active: boolean
```
ushkarev, line 51: not sure we'd wanna keep `active` here long term cos if a role is not allowed, we can just not list it inside `IncidentTypeConfiguration.prisonerRoles `

ushkarev (approved):

server/data/incidentTypeConfiguration/types.ts
xoen: Possible, not sure, but I think it makes sense to keep it at the moment as we're effectively translating the NOMIS config and the alternative is to remove these inactive roles altogether which is probably a step too far.

PR #189, opened 2024-09-25
[IR-511] Added functionality to use conditional fields as part of the form wizard

jonesj2193:


server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -121,20 +132,27 @@ declare module 'hmpo-form-wizard' {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       validateFields(req: FormWizard.Request, res: Express.Response, callback: (errors: any) => void)
 
-      locals(req: Request, res: Express.Response, next: Express.NextFunction): Promise
+      // eslint-disable-next-line no-underscore-dangle
+      _locals(req: Request, res: Express.Response, next: Express.NextFunction): Promise
+
+      locals(req: Request, res: Express.Response, next: Express.NextFunction): object
```
xoen, line 138: Why there are "two" locals now?

server/data/frontendComponentsClient.ts
```diff
@@ -10,6 +10,33 @@ export interface Component {
...
+  caseLoads: CaseLoad[]
+  services: Service[]
+}
+
+export interface FrontendComponentsResponse {
+  header?: Component
+  footer?: Component
+  meta: FrontendComponentsMeta
+}
+
```
xoen, line 39: I'm confused, is this FrontendComponent stuff needed for the conditional fields? 🤔

server/routes/generic/fields.ts
```diff
@@ -499,6 +488,30 @@ const fields = {
       },
     ],
   },
+  whereOutsideHospital: {
```
xoen, line 491: Is this the infamous "or [here a bunch of nested answers]" as an example :)?

May I suggest we only use this conditional fields functionality with the `dateRequired`/`commentRequired` so that we don't accidentally set the wrong expectations? (as we know questions/answers nested within answers opens a whole can of worms)

(but I misunderstood this, disregard this comment ☺️)

server/helpers/field/renderConditionalFields.ts
```diff
@@ -0,0 +1,41 @@
...
+        const components = conditionalFields.map(conditionalFieldKey => {
+          const conditionalField = field.prefix
+            ? fields[`${field.prefix}[${conditionalFieldKey}]`]
+            : fields[conditionalFieldKey as string]
+
+          if (!conditionalField) {
+            return undefined
+          }
+
+          return req.services.frontendComponentsService.getComponent(conditionalField.component, conditionalField)
```
xoen, line 30: Ah the components for the fields are coming from Frontend Component Service 👌 

Does it have to be in `req.services`?
I do wonder if this could use the same mechanism we use the same APIs mechanism we use for stuff like the IncidentReporting/OffenderSearch API etc  🤔.

server/controllers/base/formInitialStep.ts
```diff
@@ -74,4 +129,70 @@ export default class FormInitialStep extends FormWizard.Controller {
   formError(fieldName: string, type: string): FormWizard.Controller.Error {
     return new FormWizard.Controller.Error(fieldName, { args: {}, type, url: '/' })
   }
+
+  setupDateInputFields(fields: FormWizard.Fields, errorlist: FormWizard.Controller.Error[]): FormWizard.Fields {
```
ushkarev, line 133: we already had some date inputs previously, what's this new function intended to do on top of what already happened before?

server/@types/hmpo-form-wizard/index.d.ts
jonesj2193: So these are built in functions within the form wizard. From what I can tell and understand, the `_locals` is used when there are async processes going on within it, hence the promise at the end, whereas the normal `locals` is simpler and is used when you don't need async processes within it.

server/data/frontendComponentsClient.ts
jonesj2193: I added these in along with the `getComponents` as they are used in the service I added which is then used as part of generating the conditional fields.

server/routes/generic/fields.ts
jonesj2193: Yes this is that case. It is simply in here as it was the example I was previously trying to recreate when discussing the generic stuff with Jonathon. 
As we discussed yesterday, we will be looking to only implement text and date inputs as they are required and defined by the NOMIS config. But the functionality is the same, so we will do so because of data management issues etc, but it's there to implement in future if they insist when it comes to new designs.

server/routes/addPrisoner/fields.ts
```diff
@@ -13,8 +13,8 @@ const fields = {
     },
     items: [
       {
-        text: 'Please select one',
-        selected: true,
+        text: 'overwritten during runtime',
```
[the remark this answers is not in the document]
jonesj2193: These are set in the controllers. So I have a function in the controller `setOptions` which imports the values from the constants in the report configurations you made so that we don't need to manually keep the field options maintained. It reads them in from there and sets the options as part of the page being generated.

server/routes/generic/fields.ts
```diff
@@ -499,6 +488,30 @@ const fields = {
...
+    component: 'govukRadios',
```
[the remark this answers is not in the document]
jonesj2193: So same as the response to Aldo's comment elsewhere above. The functionality regardless of which component type is the same. This was just an example I was trying to get working as it was included in one of the prototype screens Jonathon had made. Which it is now working, but we aren't going to be actively using it for these types. We will just be looking to implement text and date conditional fields, but the underlying functionality is the same either way. This was just kind of a proof of concept.

server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -70,9 +71,10 @@ declare module 'hmpo-form-wizard' {
     NotSpecified = 9,
   }
 
-  function FormWizard(steps: Steps, fields: Fields, config: FormWizardConfig)
+  function FormWizard(steps: FormWizard.Steps, fields: FormWizard.Fields, config: FormWizardConfig)
 
   namespace FormWizard {
```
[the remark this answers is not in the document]
jonesj2193: Yeah I copied the changes to the form wizard and types etc from the Locations repo. I spoke to them and they said they have done a lot of work and made changes since working with it more. I also thought these updated and more vigorous types will come in useful when we try to automate their creation as I am look at now.

server/controllers/base/formInitialStep.ts
jonesj2193: So I did mean to leave this out actually, so I can go back and remove it. However, they used it as it related to the gov date input fields, where it splits day, month and year into separate input boxes. They are rather fiddly to deal with and validate etc. I tried playing about with it when I first was bringing in the form wizard, they have added a bunch of functions including validation checks etc to deal with them. They were told they had to use the date inputs as part of their designs, if we can use the date picker instead (as we have done so far) it makes our lives easier and I don't think we need these functions. So happy to take it out or can leave in in case we need it in future.

server/data/frontendComponentsClient.ts
ushkarev: the conditional fields stuff doesn't use `getComponents`

server/@types/hmpo-form-wizard/index.d.ts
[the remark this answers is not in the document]
ushkarev: oh absolutely! i love stricter, more accurate types! would be good to update the comment at the top to say where it came from

server/controllers/base/formInitialStep.ts
ushkarev: i would prefer to leave it out for now because we don't have tests for it and we don't know if we will ever use it should UCD go down the route of date pickers

server/middleware/addServicesToRequest.ts
```diff
@@ -0,0 +1,10 @@
+import { RequestHandler } from 'express'
+import { Services } from '../services'
+
+export default function addServicesToRequest(services: Services): RequestHandler {
```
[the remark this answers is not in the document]
jonesj2193: It was used and put in so that it's easier to access them in the controllers and functions that have the req parsed into them. But we don't need to use it.

server/services/frontendComponentsService.ts
```diff
@@ -0,0 +1,38 @@
+import nunjucks from 'nunjucks'
+import FrontendComponentsClient, { AvailableComponent } from '../data/frontendComponentsClient'
+import kebabCase from '../formatters/kebabCase'
+
+export default class FrontendComponentsService {
```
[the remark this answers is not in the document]
jonesj2193: The service is used in `renderConditionalFields.ts:30`

server/helpers/field/renderConditionalFields.ts
jonesj2193: I have mentioned this with Igor in a separate chat. I put it in a service as there are additional things being done as part of the functions and the api call defined in the client is used in there. Thought it may make sense to keep the client for dealing with the api calls and then having the services uses them and adds whatever additional functionality. But can make changes if preferred.

ushkarev: but this line does not call the frontend components client or api. it calls the service method which renders a nunjucks template.

server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -1,7 +1,8 @@
 // Copied from https://github.com/ministryofjustice/hmpps-strengths-based-needs-assessments-ui/blob/c1d2e0ac64ce2be0ea0a27919cd34f8882b2d5c0/server/%40types/hmpo-form-wizard/index.d.ts
```
ushkarev, line 1: could you replace this comment with the reference to wherever the type updates come from pls?

the latest ones in `hmpps-strengths-based-needs-assessments-ui` are quite different.

server/@types/express/index.d.ts
```diff
@@ -25,6 +26,7 @@ export declare global {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
+      services?: Services
```
ushkarev, line 29: could you pls remove the `services` property on requests? i am pretty sure you'd already removed the request handler that put it there.

i think the separation of requests and shared services will be easier to reason about if services are passed to middleware _directly_, not via requests. request flows already have various instantiated api classes via `res.locals.apis`

server/@types/hmpo-form-wizard/index.d.ts
xoen: _**FWIW**_ I've noticed some types weirdness related to `Step` for example. The `hmpo-form-wizard` documentation and examples suggest pretty much all fields are optional?

[from the docs: ](https://github.com/HMPO/hmpo-form-wizard/tree/master?tab=readme-ov-file#additional-step-options)
> The minimum amount of configuration for a wizard step is the next property to determine where the user should be taken after completing a step. A number of additional properties can be defined.
> [...]

Also some of [their examples](https://github.com/HMPO/hmpo-form-wizard/tree/master?tab=readme-ov-file#usage) have steps with nothing in them, e.g.

```
// steps.js
module.exports = {
  '/step1': {
    next: 'step2'
  },
  // [...]
  '/step4': {},
  '/not-old-enough': {}
}
```

🤷

server/data/frontendComponentsClient.ts
jonesj2193: I forgot to remove these bits from the client, I can do still if wanted.

server/routes/generic/fields.ts
jonesj2193: This has been removed now

server/helpers/field/renderConditionalFields.ts
jonesj2193: this has been changed as discussed

server/services/frontendComponentsService.ts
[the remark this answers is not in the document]
jonesj2193: removed as discussed

server/middleware/addServicesToRequest.ts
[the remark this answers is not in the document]
jonesj2193: this has been removed

server/routes/generic/fields.ts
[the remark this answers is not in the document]
jonesj2193: this has been removed now

server/controllers/base/formInitialStep.ts
jonesj2193: this has been removed now

server/@types/hmpo-form-wizard/index.d.ts
jonesj2193: Updating the link now. 
In terms of the types, yeah it is a bit strange in that aspect as I believe it is very flexible as most if not all fields are not necessarily needed to actually generate and show something, but then lots can be added for additional configurability.

server/@types/express/index.d.ts
jonesj2193: Oh yeah missed this, I will go back and remove it

ushkarev (approved):

server/data/frontendComponentsClient.ts
ushkarev: this is totes fine actually. even though it's not part of the changes in this pr, i noticed that our old way of retrieving the header/footer components is 1) deprecated and 2) this alternative also brings in caseloads (which we will need one way or another)

PR #191, opened 2024-09-26
IR-68: Remove form wizard's session timeout check

ushkarev:
…because this is already handled by other middleware.

Relates to #167 and #181


jonesj2193 (approved):
LGTM

PR #197, opened 2024-09-28
IR-500: Tweaks to prisoner search

ushkarev:


server/middleware/populateIncident.ts
```diff
@@ -1,23 +1,18 @@
...
+      const { incidentReportingApi } = res.locals.apis
       res.locals.incident = await incidentReportingApi.getReportById(req.params.id)
     } catch (error) {
       logger.error(error, 'Failed to populate incident')
       next(error)
+      return
     }
 
-    return next()
+    next()
```
ushkarev, line 16: when catching an error, the next request handler was called twice: `next(error); next()`

server/middleware/populateReport.ts
```diff
@@ -0,0 +1,19 @@
+import type { NextFunction, Request, Response } from 'express'
+
+import logger from '../../logger'
+
+// eslint-disable-next-line import/prefer-default-export
+export function populateReport() {
+  return async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+    try {
+      const { incidentReportingApi } = res.locals.apis
+      res.locals.incident = await incidentReportingApi.getReportById(req.params.id)
```
ushkarev, line 10: i think we need to call reports “reports”, because incidents are actually _groups_ of reports

jonesj2193 (approved):
LGTM

PR #204, opened 2024-10-04
Generate fields

jonesj2193:
Created function to generate all fields for a given incident type. Added a page using the form wizard to display generated fields


server/services/frontendComponentsService.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  getComponent(macroName: string, params = {}) {
+    const macroParams = JSON.stringify(params, null, 2)
+    const filename = this.macroNameToFilepath(macroName)
+    const macroString = `
+      {%- from "${filename}/macro.njk" import ${macroName} -%}
+      {{- ${macroName}(${macroParams}) -}}
+    `
+    return nunjucks.renderString(macroString, undefined)
+  }
```
ushkarev, line 37: why did you want to bring this back? the purpose of “frontend components” is purely to load the header and footer from an *external api*. you already made a mechanism in the previous pr to call a macro by name

server/middleware/addServicesToRequest.ts
```diff
@@ -0,0 +1,10 @@
+import { RequestHandler } from 'express'
+import { Services } from '../services'
+
+export default function addServicesToRequest(services: Services): RequestHandler {
+  return async (req, _res, next) => {
+    req.services = services
+
+    next()
+  }
+}
```
ushkarev, line 10: we don't need this. in fact, you don't use anything from `res.services`

server/@types/express/index.d.ts
```diff
@@ -25,6 +26,7 @@ export declare global {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
+      services?: Services
```
ushkarev, line 29: we don't need this. in fact, you don't use anything from `res.services`

server/routes/generic/fields.ts
```diff
@@ -487,6 +488,30 @@ const fields = {
       },
     ],
   },
+  whereOutsideHospital: {
```
ushkarev, line 491: as suggested previously, could we leave demos of conditional functionality to purely a single comment box or a single date input?

PR #208, opened 2024-10-07
Generate fields

jonesj2193:
Created function to generate all fields for a given incident type. Added a page using the form wizard to display generated fields


xoen:
Closing this one to avoid confusion, as it's the base for PR #209 which is this + some fixes to paths, steps, etc...

PR #209, opened 2024-10-07
Generate steps

jonesj2193:
Added functions and example path to generate steps from incident config.
This is branched off of the fields branch. So a lot of the diff initially comes from that PR. Once that is merged in, the diff will reduce in this PR.


server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -1,27 +1,36 @@
...
         fields: [question.id],
-        template: 'index',
+        controller: QuestionsController,
+        template: 'questionPage',
       }
-    }
-  })
-
-  steps[`/${config.startingQuestionId}`].entryPoint = true
-  steps['/'] = steps[`/${config.startingQuestionId}`]
```
ushkarev, line 24: i preferred this tbh. i don't love the idea of an initial auto-skipped step

ushkarev (approved):

PR #210, opened 2024-10-09
IR-534: Recreate HMPO form wizard types afresh

ushkarev:
…with clear separation between library and custom additional properties. It’s better to use declaration merging.

Types copied from `hmpps-locations-inside-prison` & `hmpps-strengths-based-needs-assessments-ui` include a lot of application-specific properties that are not applicable to incident reporting and associated code does not exist in this repository.

This PR is easiest to review by first looking at changes made to application code: nearly none! Some types have been tightened up, other declarations now specify types rather than using type assertions (i.e. types are checked rather than assumed).

In future, we will need to tidy up custom properties hanging off form wizard types:
- reuse existing functionality (back links & conditional fields are already supported by the library)
- custom field properties should be nested into sub-objects to not pollute the namespace

### DIFF
--- a/server/data/incidentTypeConfiguration/formWizard.ts
+++ b/server/data/incidentTypeConfiguration/formWizard.ts
@@ -1,5 +1,5 @@
 import FormWizard from 'hmpo-form-wizard'
-import { IncidentTypeConfiguration } from './types'
+import type { IncidentTypeConfiguration } from './types'
 import QuestionsController from '../../controllers/wip/questionsController'
 
 // TODO: Add tests once steps structure is more stable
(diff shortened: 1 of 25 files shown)


server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -1,5 +1,5 @@
 import FormWizard from 'hmpo-form-wizard'
-import { IncidentTypeConfiguration } from './types'
+import type { IncidentTypeConfiguration } from './types'
```
xoen, line 2: 👌

ushkarev: sorry, i kno these keywords are trivial and make no difference to the output but i cannot help myself 😅

PR #213, opened 2024-10-10
IR-505: Warn if an active question has no active answers

ushkarev:


server/data/incidentTypeConfiguration/validation.ts
```diff
@@ -50,12 +50,15 @@ function checkUnreachableQuestions<Q>(configGraph: Graph<Q>, dfsResult: DfsResul
   }
 }
 
-function buildConfigGraph(config: IncidentTypeConfiguration): Graph<string> {
+function buildConfigGraph(config: IncidentTypeConfiguration, errors: Error[]): Graph<string> {
```
xoen, line 53: I'm not so sure populating errors is a concern of the function that build the graph...it doesn't feel right (if you know what I mean).

I appreciate we already loop the questions and have an `activeAnswers` variable in here...but it still doesn't feel right?

I'd almost prefer for this check to be done separately even if it means looping through the config's questions/answers another time...but I don't think performance is so much of a problem in this case, having separate validation checks for related things is possibly more important?

Don't know what you think. If you have strong opinions on this I'm happy to bow and make this go though tho.

server/data/incidentTypeConfiguration/validation.ts
```diff
@@ -34,15 +35,26 @@ function checkStartingQuestion(config: IncidentTypeConfiguration, errors: Error[
   }
 }
 
-function checkUnknownQuestions<Q>(configGraph: Graph<Q>, errors: Error[]) {
+function checkQuestionsWithoutAnswers(config: IncidentTypeConfiguration, errors: Error[]): void {
+  Object.values(config.questions)
+    .filter(question => question.active === true)
```
xoen, line 40: Our config format should be relatively good quality, so maybe  `.filter(question => question.active)` would be fine?

The type of active is `boolean` and typecheck should throw an error if that is not the case no?

ushkarev: this is copied from *your* code

xoen: Still, what do you think 🤔?

xoen (approved):

PR #219, opened 2024-10-11
IR-533: Install `getFromContext(…)` function to look up values from template context by .-separated key path

ushkarev:
**Don’t merge: might not be needed**

Inspired by `ctx()` from [`hmpo-components`](https://github.com/HMPO/hmpo-components/blob/01473508284e1d06ebc72585f970f25f17fba49d/lib/locals.js#L59) but has access to the entire template context, not just `res.locals`.

This is heavily used in HMPO components library and will be helpful for use with HMPO form wizard & templates. Example:

```nunjucks
A simple case where the parameter path is known in advance:
<input name="firstName" value="{{ getFromContext('values.firstName') }}" />

…but the more complex situation is when the parameter path to look up is in a variable:
{% for fieldName in fields %}
  <input name="{{ getFromContext(fieldName) }}" value="{{ getFromContext('values.' + fieldName) }}" />
{% endfor %}
```

### DIFF
--- a/server/utils/nunjucksSetup.test.ts
+++ b/server/utils/nunjucksSetup.test.ts
@@ -1,5 +1,6 @@
 import express from 'express'
 import nunjucks from 'nunjucks'
+import request from 'supertest'
 
 import nunjucksSetup from './nunjucksSetup'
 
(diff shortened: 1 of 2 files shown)


ushkarev:
unlikely to need this: only top-level context variables are inaccessible without this and we're unlikely to flood it with fields

PR #220, opened 2024-10-13
IR-68: Have test application preserve existing `res.locals`

ushkarev:


server/routes/testutils/appSetup.ts
```diff
@@ -42,15 +42,15 @@ function appSetup(services: Services, production: boolean, userSupplier: () => E
   nunjucksSetup(app)
   app.use((req, res, next) => {
     req.user = userSupplier()
-    res.locals = {
+    Object.assign(res.locals, {
```
xoen, line 45: I think the point of using `Object.assign()` was to preserve any existing properties? But I doubt it makes much difference in this case?

xoen (approved):

server/routes/testutils/appSetup.ts
ushkarev: any prior middleware could set locals and they'd be lost so i don't think it's a good idea to completely replace what's there

xoen: Sorry when I looked at your PR earlier I read it the other way around! 🤦‍♂️

PR #223, opened 2024-10-14
Moved field rendering logic into a partial

xoen:
The `renderField` partial can now be used instead of using `callAsMacro()` (which may invoke any macro based on the `field.component` property). This should be more explicit.

**NOTE**: This will most definitely need tweaks as there are some questions regarding how the conditional "fields" (which are not FormWizard fields at the moment) would be processed (validation/submission/etc). Furthermore answers don't have an id and the ids/names for these conditional fields are currently generated from the question id/answers code/(comment|date) which will need some more thinking.


server/views/pages/wip/questions/questionPage.njk
```diff
@@ -57,9 +51,9 @@
         {% endif %}
 
         {% block fields %}
-          {% for key, fieldsOptions in fields %}
-            {% if fieldsOptions.component and not fieldsOptions.skip %}
-              {{ callAsMacro(fieldsOptions.component)(fieldsOptions) }}
+          {% for _fieldId, field in options.fields %}
+            {% if field.component and not field.skip %}
```
ushkarev, line 55: `field.skip` is not a param that we have defined… i don't think it will do anything

xoen: I'll remove that bit then 👌

xoen: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/223/commits/d2571e5fd83c3ee54260d014147a1950fb88a208

server/views/components/wip/renderField/macro.njk
```diff
@@ -0,0 +1,80 @@
...
+            }
+          },
+          items: items
+      }) }}
+  {% else %}
+    <strong>WARNING: Unknown field's component {{ field.component }}
+  {% endif %}
+{% endmacro %}
+
+{% macro conditionalHtml(item) %}
```
ushkarev, line 48: this macro uses variables (`field`, `formValues`, `conditionalHtml`) internally that haven't been passed to it so i am pretty certain it will throw an exception.

ushkarev (approved):

server/views/components/wip/renderField/macro.njk
xoen: as mentioned offline...nunjucks being nunjucks it just swallows the issue and silently returns `undefined`

xoen: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/223/commits/78498059f118346f543d812d1515d3ac83670acf

jonesj2193:
A general question on this PR, why has the logic been changed and moved regarding the conditional fields stuff when there are already functions that have been added to generate those fields and helps keep the corresponding nunjucks templated simpler than they now seem to be? Also if there is a note about their corresponding ids etc when that wasn't an issue before as the fields were created and generated and called in so that they could be referred to?

ushkarev:
the main thing is that the form wizard library already has built-in support for conditional field handling and validation (via the `dependent` property). this is totally a work-in-progress PR, however! in fact, i don't imagine much of the nunjucks code will remain. i have a separate local branch that demos the minimum that's needed and that came out to much less code.

xoen:
Hi Josh, the existing code had a number of issues (that is fine as it was a spike right?). For example there is a lot of logic that seem to be reimplemented but it is already provided by Form Wizard.

For example, the logic for the conditional fields you refer to [assumed that only one between comment/date was allowed](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/main/server/data/incidentTypeConfiguration/wip/generateFieldsWithConditionals.ts#L10).

> Also if there is a note about their corresponding ids etc when that wasn't an issue before as the fields were created and generated [...]

The existing logic assumed/worked on the assumption that only one answer would have a comment and/or date but this is not the case for checkboxes. This is why [a name like `${questionId}Comment` wouldn't work](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/69afe7468193e2cbb2d508b0ca8875337c457066/server/data/incidentTypeConfiguration/wip/generateFieldsWithConditionals.ts#L17) and that's what the note you're referring to is about (I tried to explain this in [this commit message](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/223/commits/dcbd8958eae9c350360950660fa787c03a3ea745)).

> [...] there are already functions that have been added to generate those fields and helps keep the corresponding nunjucks templated simpler than they now seem to be?

We can argue whether the `callAsMacro()` makes the template simpler however - and leaving aside how `callAsMacro()` is implemented - this makes the logic quite opaque. Furthermore it forces the fields to have a lot of markup which is a presentation concern rather than a field concern (which in my mind should mainly deal with validation rather than presentation like it was the case in the version of the fields you're referring to).

As mentioned before we're trying to gradually add things to control what we add and also to better understand what things do. We should aim to use functionality provided by Form Wizard as much as possible but the spikes have a lot of logic which seemed to be already provided by the library.

Igor is also playing with it but we should have a chat in case any of this is not clear.

ushkarev (approved):

PR #225, opened 2024-10-15
Added answers IDs to DPS config

xoen:
NOMIS has a concept of "answer id" which we didn't copy. 

This is very useful when building forms inputs as we can use these IDs to generate unique names
for things like comment/date inputs.

Forms can use these IDs to generate unique names for input that we can use to discriminate submitted values.

### For example:

**Question (id: '12345'): Which part were injured?**
  - [X] Answer (id: 'A1'): Neck
    - Comment required (input name: "12345-A1-comment"): `[ lorem ipsum ]`
  - [X] Answer (id: 'A2'): Leg
    - Comment required (input name: "12345-A2-comment"): `[ lorem ipsum ]`

**Question (id: '6789'): Was it a serious injury?**
  - [ ] Answer (id: 'A1'): Yes
    - Comment required (input name: "6789-A1-comment"): `[ lorem ipsum ]`
  - [ ] Answer (id: 'A2'): No
    - Comment required (input name: "6789-A2-comment"): `[ lorem ipsum ]`

(in this example the answer ids are only locally unique, in NOMIS answers' ids are globally unique. Our config would use these IDs but in practice it shouldn't matter as long as the names for these comment/date fields are unique and can be used to determine what question/answer they refer to)

### DIFF
--- a/server/data/incidentTypeConfiguration/types.ts
+++ b/server/data/incidentTypeConfiguration/types.ts
@@ -27,7 +27,7 @@ export interface QuestionConfiguration {
 }
 
 export interface AnswerConfiguration {
-  /** Answer identifier */
+  /** Answer ID, useful to generate unique names for comment/date inputs */
   id: string
   /** Answer as seen by machines, e.g. it shouldn't change.
    *
--- a/server/data/incidentTypeConfiguration/types.ts
+++ b/server/data/incidentTypeConfiguration/types.ts
@@ -27,6 +27,8 @@ export interface QuestionConfiguration {
 }
 
 export interface AnswerConfiguration {
+  /** Answer identifier */
+  id: string
   /** Answer as seen by machines, e.g. it shouldn't change.
    *
    * For example 'DRONE RECOVERY' */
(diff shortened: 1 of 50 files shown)


xoen:
@ushkarev this is one of the changes we discussed earlier which should solve the issue of the comment/date input names being unique even when multiple options can be selected (checkboxes)

server/data/incidentTypeConfiguration/types.ts
```diff
@@ -27,6 +27,8 @@ export interface QuestionConfiguration {
 }
 
 export interface AnswerConfiguration {
+  /** Answer identifier */
```
ushkarev, line 30: I think we ought to say that this is internal and only to be used for distinguishing and linking form fields. And that this value won't be sent to the api nor downstream data consumers

ushkarev (dismissed):

server/data/incidentTypeConfiguration/types.ts
xoen: Aren't ids always internal? 🤔 I guess this is not going to be sent to the API, or at least not at the moment.

I'll try to think of some wording that makes sense.

ushkarev: quite often you’d use an “id” to talk with an api so i just wanna make sure we remember that we won’t in this case :)

ushkarev (approved):

PR #226, opened 2024-10-16
IR-68: Nunjucks utilities to help keep presentation logic within template files

ushkarev:
The `govukCheckedItemsConditional` filter is used to add conditional html to GDS radio or checkbox components.

```typescript
res.render('template', {
  items: [
    { text: 'A', value: 'a' },
    { text: 'B', value: 'b' },
  ],
})
```

```nunjucks
{% set conditional %}
  {{ govukInput({
    name: 'comment',
    id: 'comment',
    label: { text: 'Comment' }
  }) }}
{% endset %}
{{ govukCheckboxes({
  id: 'sample',
  name: 'sample',
  fieldset: { legend: { text: 'Sample' } },
  items: items | govukCheckedItemsConditional({
    value: 'b',
    html: conditional
  })
}) }}
```

---

The `govukCheckedItemsDivider` filter adds a divider to GOVUK checkbox or radio component items.

```nunjucks
{{ govukCheckboxes({
  id: 'sample',
  name: 'sample',
  fieldset: { legend: { text: 'Sample' } },
  items: items | govukCheckedItemsDivider()
}) }}
```

---

The `panic` tag is used to unconditionally stop template rendering – useful to ensure we never reach an invalid condition.

```nunjucks
{% if expectedCondition1 %}
  render something
{% elif expectedCondition2 %}
  render something else
{% else %}
  {% panic 'this should never happen' %}
{% endif %}
```

### DIFF
--- a/server/utils/govukFrontend.test.ts
+++ b/server/utils/govukFrontend.test.ts
@@ -7,6 +7,7 @@ import {
   govukSelectSetSelected,
   govukCheckedItems,
   govukMultipleCheckedItems,
+  govukCheckedItemsConditional,
 } from './govukFrontend'
 
 describe('findFieldInGovukErrorSummary', () => {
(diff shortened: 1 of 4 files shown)


xoen (approved):
I'm not too keen on the custom nunjucks extension, especially for something so simple but nunjucks is...the way it is, so whatever :)

PR #229, opened 2024-10-18
IR-531: Validate & format time

ushkarev:


server/controllers/index.ts
```diff
@@ -46,4 +51,30 @@ export abstract class BaseController extends FormWizard.Controller {
...
   }
+
+  csrfGenerateSecret(req: FormWizard.Request, res: Express.Response, next: Express.NextFunction): void {
+    // copy application middleware CSRF token into form wizard for sanity
+    req.sessionModel.set('csrf-secret', res.locals.csrfToken)
+    next()
+  }
 }
+
+Object.assign(FormWizard.Controller.validators, {
```
ushkarev, line 62: unfortunately, the validators set is not read from an _instance_ of a controller so we have to add to the global set

server/controllers/index.ts
```diff
@@ -47,3 +52,23 @@ export abstract class BaseController extends FormWizard.Controller {
...
   }
 }
+
+Object.assign(FormWizard.Controller.validators, {
+  ukDate(value: FormWizard.Value): boolean {
+    if (value === '') return true
+    try {
+      return Boolean(parseDateInput(value))
+    } catch (e) {
+      return e === undefined
```
xoen, line 68: I'm not quite sure how the error could be undefined 🤔 - The "blank input" is already covered by the if above, is this for something else? Is this any different than `return false` when there is an exception?

ushkarev: this is gross isn't it? js requires the catch block to have a param. eslint requires it is used. i don't want to log something as boring as an incorrectly-typed date. since `e` is always going to be an error instance , this was the tidiest way i could think of to use the variable.

server/controllers/index.ts
```diff
@@ -1,6 +1,8 @@
 import type Express from 'express'
```
xoen, line 1: Random thought, we should probably rename this file to `baseController.ts` at some point.

ushkarev: hmm dunno. maybe? right now the imports look like `import { BaseController } from '../../controllers'` but is `import { BaseController } from '../../controllers/baseController'` any better?

server/controllers/index.ts
xoen: Yeah, it is a bit gross...eslint in our way :|

I am not saying this is less gross but perhaps the following is a bit less misleading and clear that we return always `false` on error?:

```
      // eslint-disable-next-line @typescript-eslint/no-unused-vars
    } catch (_e) {
      return false
    }
```

server/controllers/index.ts
xoen: > is import { BaseController } from '../../controllers/baseController' any better?
Almost call for eslint's favourite thing isn't? (the default export) - I don't find this particularly offensive but I like the idea of `BaseController` being in `baseController.ts` as it's more predictable and much easier to find/jump to it in the editor (via CMD+P and "basecon...")

But as I said, it's a bit of random thought and not technically part of this nor a _problem_ - just an observation/preference.

xoen (approved):

PR #234, opened 2024-10-21
IR-519: Various improvements to questions steps/fields

xoen:
This is starting to grow so I think it's a good idea to merge this batch of improvements:
- fix for checkboxes not going to next page
- retain user selection on error
- validate date fields with `ukDate`
- don't generate steps/fields for inactive questions/answers as it's wasteful
- added tests for `generateSteps()`/`generateFields()`
- tweaks to FormWizard types

### DIFF
--- a/server/@types/hmpo-form-wizard/index.d.ts
+++ b/server/@types/hmpo-form-wizard/index.d.ts
@@ -25,50 +25,52 @@ declare module 'hmpo-form-wizard' {
     /** Use in complex generic forms that might allow multiple values in any fields; have checkbox components */
     type MultiValues = Record<string, MultiValue>
 
+    /** Possible conditions for next steps */
+    type NextStepCondition =
+      | {
+          /** field, op and value. op defaults to '===' */
+          field: string
+          op?: '>' | '>=' | '<' | '<=' | '==' | '===' | '!=' | 'before' | 'after' | 'in' | 'all' | 'some'
+          value: Value
+          next: string
+        }
+      | {
+          /** an operator can be a function */
+          field: string
+          op: (fieldValue: Value, req: Request, res: Express.Response, con: unknown) => boolean
+          value: Value
+          next: string
+        }
+      | {
+          /** next can be an array of conditions */
+          field: string
+          value: Value
+          next: NextStep[]
+        }
+      | {
+          /** a condition can be a function specified by fn */
+          fn: (req: Request, res: Express.Response, con: unknown) => boolean
+          next: string
+        }
+      | {
+          /** a condition can be a controller method specified by name */
+          fn: string
+          next: string
+        }
+      | {
+          /** the next option can be a function to return a dynamic next step */
+          field: string
+          value: Value
+          next: (req: Request, res: Express.Response, con: unknown) => string
+        }
+      /** use a string as a default next step */
+      | string
+
     type NextStep =
       /** next can be a relative string path */
       | string
       /** next can be an array of conditions */
-      | (
-          | {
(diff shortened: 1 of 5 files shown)


server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -25,50 +25,50 @@ declare module 'hmpo-form-wizard' {
     /** Use in complex generic forms that might allow multiple values in any fields; have checkbox components */
     type MultiValues = Record<string, MultiValue>
 
+    /** Possible conditions for next steps */
+    type NextStepCondition =
```
xoen, line 29: This is just the list of valid values moved into its own type.

server/views/components/wip/renderField/_conditionalHtml.njk
```diff
@@ -21,7 +21,7 @@ USAGE:
         text: "Date"
       },
       value: values[field.name + '-' + item.id + '-date'],
-      leadingZeros: "true"
+      leadingZeros: true
```
ushkarev, line 24: Does this work now? I don't think they fixed it yet…?

https://github.com/ministryofjustice/moj-frontend/issues/780

xoen: Main reason I changed to a boolean is because when it was a string the dake picker was not rendering correctly.

Oddly, when the property is a string it throws a `Uncaught TypeError: this.config.leadingZeros.toLowerCase is not a function` (which is odd because surely a string would have the `toLowerCase()` method):

```JavaScript
Datepicker.prototype.setLeadingZeros = function () {
  if (typeof this.config.leadingZeros !== "boolean") {
    if (this.config.leadingZeros.toLowerCase() === "true") {
      this.config.leadingZeros = true;
    }
    if (this.config.leadingZeros.toLowerCase() === "false") {
      this.config.leadingZeros = false;
    }
  }
};
```

`leadingZeros: true` doesn't seem to make much difference at the moment AFAICT but at least the date picker works.

xoen:
FYI: Rebased

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -77,41 +82,92 @@ export function generateFields(config: IncidentTypeConfiguration): FormWizard.Fi
...
+ * the step specificed in the `next` property of the next step condition
+ * (`q2` in the example above)
+ *
+ * @param submittedValues values submitted by the user
+ * @param _req request not used
+ * @param _res response not used
+ * @param condition next step condition with value to match
+ * @returns true if the value submitted by the user include `condition.value`
+ */
+export function checkMultipleValues(
```
ushkarev, line 161: i think this function would be better placed outside of this file as it is not specifically tied to generated question/answer steps

xoen: it is actually, it's in the specific FormWizard format required in the next conditional steps and it also tied to how we currently generate the steps' next (it will likely change once we group answers together by next question id).

For this reason I don't wanna move this in a separate file or even make it public as it's very tied to the current steps.

ushkarev: i still think this function is not tied to anything specific here, it's type is what the form wizard lib expects as a next step condition “op”
```typescript
op: (fieldValue: Value, req: Request, res: Express.Response, con: NextStepCondition) => boolean
```
it wouldn't need any modification to work for checkboxes elsewhere. but whatevs :)

ushkarev (approved):

PR #237, opened 2024-10-22
IR-548: Create a report

ushkarev:
This adds the first two basic steps that are required before a report can be saved by the API:
- select an incident type
- add basic details

Not yet done because decisions haven’t been made how it will work:
- add the ability to select a location outside of prison/active caseload, e.g. PECS or a region
- no role checking is performed
- some content is incomplete / missing
- error messaging is bare-bones (cannot distinguish in summary which fields errors relate to)


Mjwillis (approved):
Nice one!

server/reportConfiguration/constants/types.ts
```diff
@@ -83,3 +83,8 @@ export type Type = (typeof types)[number]['code']
  * @deprecated
  */
 export type NomisType = (typeof types)[number]['nomisCode']
+
+/** Look up an incident type by code */
+export function getTypeDetails(code: string): (typeof types)[number] | null {
```
xoen, line 88: Perhaps we can have a `TypeDetails` type set as

```Typescript
export type TypeDetails = (typeof types)[number]
``` 

`Type` becomes:

```Typescript
export type Type = TypeDetails['code']
```

and this function signature becomes:

```Typescript
export function getTypeDetails(code: string): TypeDetails | null {
}
```

overkill? Besides it's a bit odd that the return type of this function here is the specific union of all the valid report types rather than a simpler

```Typescript
export type TypeDetails = {
  code: string
  description: string
  active: boolean
  nomisCode: string
}
```

but it's probably more reflective of the valid report types allowed (and it would evolve over time as more types are added or deactivated)

xoen: (just an observation that `(typeof types)[number]` is used twice and `TypeDetails` may be a nice name for it but you don't need to necessarily change anything)

server/controllers/index.ts
```diff
@@ -70,6 +70,50 @@ export abstract class BaseController<
     req.sessionModel.set('csrf-secret', res.locals.csrfToken)
     next()
   }
+
+  /**
+   * Builds an object of values for all fields and steps from session or, alternatively,
```
xoen, line 74: Isn't this what Form Wizard does anyway? 🤔 

I can see values from previous steps but I'm probably missing what this is about.

ushkarev: i found that `req.form.values` doesn’t just contain what you expect, but other things too (like CSRF token) plus it’s not necessarily valid data. i wonder if this is overkill, but retrieving things from the session model felt safer to me.

once we do more manual testing, we can maybe drop this in favour of values…?

xoen: yeah, I didn't look at this properly but it seems like it's somewhat adding something that should really be in Form Wizard 🤔

ushkarev: yeah i agree, it does feel that way. i just recall that there was some weirdness i hit during my testing.

server/views/components/wip/fields.njk
```diff
@@ -0,0 +1,15 @@
+{% from "components/wip/field.njk" import renderField %}
+
+{% macro renderFields(options, values, errors) %}
+
+  {% for fieldName, field in options.fields %}
+    {# NB:
+      - dependent fields are currently only used in conditionals so are rendered separately
+      - fields starting with _ are internally generated
+    #}
+    {% if not field.dependent and not fieldName.startsWith("_") %}
```
xoen, line 10: what's an internally generate field 🤔? Currently comment/date fields are just ignored because of the `dependant` property not being null (as in this case here)

server/routes/reports/createReport.test.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  describe('Step 2: entering date and description', () => {
+    beforeEach(() => {
+      return agent.post('/create-report').send({ type: 'DISORDER' })
+    })
+
+    const incidentDateAndTime = new Date('2024-10-21T16:32:00+01:00')
+    const validPayload = {
+      incidentDate: '21/10/2024',
+      '_incidentTime-hours': '16',
```
xoen, line 98: I'm not so sure about this naming convention, these are not private fields or anything, but I guess maybe there is some value in distinguishing them from "normal" fields? Unless we want to render them separately from the usual `for (field of fields)` loop

server/routes/reports/createReport.ts
```diff
@@ -0,0 +1,226 @@
+// eslint-disable-next-line max-classes-per-file
```
xoen, line 1: There is a lot in this file... :)

server/views/components/wip/checkedItemsField.njk
```diff
@@ -0,0 +1,68 @@
+{% from "govuk/components/checkboxes/macro.njk" import govukCheckboxes %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+
+{% macro renderCheckedItemsField(fieldName, field, options, values, errors) %}
```
xoen, line 4: Why the complicated name? This is rendering a field no? I think it should be `renderField()` unless I'm missing something (and yes it checks if the field is one of the ones we support but otherwise it's rendering a field no?)

server/views/components/wip/fields.njk
ushkarev: the “time” field is split into hours and minutes so in reality it is 2 fields, but we cannot count those in the loop as the parent is what drives the template. this is roughly how the hmpo library does it too.

server/routes/reports/createReport.test.ts
ushkarev: we can't render them separately: it's the owning `incidentTime` field that is used in the template

server/views/components/wip/checkedItemsField.njk
ushkarev: there are 5 field templates now and they need to be imported into one template, if they all have the same name that wouldn't work. i prefer things to have 1 official name and for imports to not rename things. (similar to how we both try to avoid default exports in typescript)

server/routes/reports/createReport.ts
```diff
@@ -0,0 +1,226 @@
...
+      logger.error({ err: e }, 'Report could not be created: %j', e)
+      const err = this.convertIntoValidationError(e)
+      // TODO: find a different way to report whole-form errors rather than attaching to specific field
+      this.errorHandler({ incidentDate: err }, req, res, next)
+    }
+  }
+
+  getNextStep(req: FormWizard.Request<CreateReport, Step2>, res: Express.Response): string | undefined {
+    // if a report was successfully created, redirect to it
+    if ('createdReport' in res.locals && res.locals.createdReport) {
```
xoen, line 206: Equivalent to this no?

```JavaScript
if (res.locals?.createdReport) {
  // ;)
}
```

ushkarev: except this makes eslint happy

server/views/components/wip/checkedItemsField.njk
xoen: I may be a bit slow here but the names are currently confusing:
- template imports `renderFields()` from `components/wip/fields.njk` 🙃 
- `fields.njk` contains `renderFields()`, uses `renderField()`
-  `field.njk` contains `renderField()`
- `field.njk` uses `renderCheckedItemsField()` which renders either checkboxes or radios 🙃 

May I suggest perhaps we keep filenames/macro names aligned? I'm happy to stick with the `renderThing()` convention here, but I'd also name the files `renderThing.njk` so it's predictable to find those (e.g. when jumping around in the editor or otherwise).

The `renderCheckedItemsField()` confused me, specifically the "checked" part. Maybe a better name would be `renderRadiosOrCheckboxes()`?  🤔 (and yes both have items which is why they're rendered by the same macro)

server/routes/reports/createReport.ts
xoen: I, for one, _**NOT**_ welcome our new robot overlords...

server/views/components/wip/dateField.njk
```diff
@@ -0,0 +1,18 @@
...
+
+  {{ mojDatePicker({
+    name: fieldName,
+    id: fieldName,
+    label: {
+      text: field.label,
+      classes: "govuk-label--m"
+    },
+    hint: { text: field.hint } if field.hint else undefined,
+    leadingZeros: "true",
```
xoen, line 13: I know we kinda had this conversation before but I still see this error when `leadingZeros` is a string:

```
Uncaught TypeError: this.config.leadingZeros.toLowerCase is not a function
```

And as a result the fancy date picker doesn't show (just the text input)

server/views/partials/formWizardLayout.njk
```diff
@@ -0,0 +1,47 @@
+{% extends "partials/layout.njk" %}
```
xoen, line 1: Nice one

server/views/components/wip/dateField.njk
ushkarev: it _totally_ does not work in _this_ template if u make it a boolean

server/views/components/wip/checkedItemsField.njk
ushkarev: mind if i do the template/macro renaming in a future pr? cos i have 2 more lined up and dread the idea of rebasing 🤣

server/routes/reports/createReport.ts
xoen: I think we may want to structure this a bit better somehow.

Mind you this works, it's just a bit chaotic with all these heterogeneous but related things all in a single file 🤔

I wonder if we could introduce a `formWizards` directory alongside routes/views etc...with a directory per view, e.g. `createReport/` and then in this directory we have files with names like `steps.ts`/`fields.ts`/`controllers.ts` and then this file just uses these and it's much smaller.

Thoughts? It doesn't have to happen as part of this PR but I like the idea unless you think it's rubbish. But think about it, this is a relatively simple wizard and the file is already this long, imagine once we add staff/prisoner involvement and what not 😬

ushkarev: yup, i agree. this file has grown a fair bit since i started it! mostly because of special date/time handling and validation.

there will be a fair bit of interaction between the stuff in here and the update report version, so it's quite unclear to me what a good structure would look like

server/views/components/wip/dateField.njk
xoen: So, as discussed offline, I was getting this error not sure why...a fresh `nvm use && rm -rf node_modules && npm install` somehow solved the problem (even tho I'm pretty sure I've run an npm install _**very recently**_)...so perhaps npm using an older version somehow?

Odd but at least this works (although my preference would be to rather have this as a boolean _**if it worked**_ 😬 )

xoen (approved):

server/views/components/wip/checkedItemsField.njk
ushkarev: hopefully this is better?
https://github.com/ministryofjustice/hmpps-incident-reporting/pull/239/commits/a29b558e5f202901e47491ec4ba95f5ab8dde9cc

server/controllers/index.ts
ushkarev: i remembered one thing! values aren't removed once set (because we don't have invalidation rules) meaning if you choose a radio button and add a conditional comment, then go back and choose a different radio button – the comment remains in `values`. that’s why i read everything from the session model (as in HMPO examples) and then manually delete unnecessary conditionals

PR #239, opened 2024-10-23
IR-531: Merge field templates

ushkarev:
- field items do not have individual `name`s in GOV.UK radio or checkbox components (or html), so these have been removed
- field items _can_ have `id`s, but for now it looks like auto-generating them in the template is tidier – it makes the generated fields structures smaller; they can be reinstated easily if it proves necessary to have stable references to items on screen (for integration tests, the value or label are better)
- radio and checkbox components’ conditional HTML is additive, i.e. there can now be multiple dependent fields for one item
- fix for radio and checkbox wrapper ids – the GOV.UK components do not expose an `id` parameter so it must be passed as a custom attribute


server/utils/govukFrontend.test.ts
```diff
@@ -123,6 +123,19 @@ describe('govukCheckedItemsConditional', () => {
     ])
   })
 
+  it('should append content if item already has conditional html', () => {
+    const itemsWithConditionalHtml: GovukRadiosItem[] = [
+      { text: 'A', value: 'a', conditional: { html: 'Comment on A' } },
+      { text: 'B', value: 'b', conditional: { html: 'Comment on B' } },
+    ]
+    expect(
+      govukCheckedItemsConditional(itemsWithConditionalHtml, { value: 'b', html: '<strong>info</strong>' }),
```
xoen, line 132: we're still renaming this macro right :P?

xoen (approved):
Beautiful :)

PR #241, opened 2024-10-24
IR-519: Improve next step conditions

xoen:
If multiple answers lead to the same next question they'll share the same next step condition.

For example:
- Q1 has 3 answers
- Answer 1 and 3 leads to Q2
- But Answer 2 leads to Q3

In this case Answer 1 and 3 would have the same next step condition.

For questions with a lot of answers leading to the same next question this means a much more compact steps object.

Also improved accuracy of some of the related Form Wizard types.

### DIFF
--- a/server/@types/hmpo-form-wizard/index.d.ts
+++ b/server/@types/hmpo-form-wizard/index.d.ts
@@ -25,20 +25,34 @@ declare module 'hmpo-form-wizard' {
     /** Use in complex generic forms that might allow multiple values in any fields; have checkbox components */
     type MultiValues = Record<string, MultiValue>
 
+    /** Operator function, returns true if submitted values matches condition */
+    type OperatorFunction = (
+      submittedValue: Value | MultiValue,
+      req: Request,
+      res: Express.Response,
+      condition: unknown,
+    ) => boolean
+
     /** Possible conditions for next steps */
     type NextStepCondition =
       | {
           /** field, op and value. op defaults to '===' */
           field: string
-          op?: '>' | '>=' | '<' | '<=' | '==' | '===' | '!=' | 'before' | 'after' | 'in' | 'all' | 'some'
-          value: Value
-          next: string
-        }
-      | {
-          /** an operator can be a function */
-          field: string
-          op: (fieldValue: Value, req: Request, res: Express.Response, con: unknown) => boolean
-          value: Value
+          op?:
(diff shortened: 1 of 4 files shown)


server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -127,12 +113,53 @@ export function generateFields(config: IncidentTypeConfiguration): FormWizard.Fi
...
+ *
+ * Answers leading to the same question are grouped together.
+ *
+ * @param question
+ * @param answers
+ * @returns next step conditions
+ */
+function nextSteps(question: QuestionConfiguration, answers: AnswerConfiguration[]): FormWizard.NextStepCondition[] {
+  // Group answers by next question id
+  const groupedByNextQuestion: Map<string | null, AnswerConfiguration[]> = new Map()
```
xoen, line 127: I was using a plain object `{}` here but tests highlighted the fact that when then iterating over its keys/values the key was converted to `string`.

Beside from being surprising this is bad in our case because a `nextQuestionId` (used to group answers) **can** be `null` and we don't want this to be converted to `'null'` which would be incorrectly interpreted as "go to the question with id 'null'" which is wrong.

Side rant: this is also why below there is a for loop instead of using `map()` on the return value of `.entries()`: That value doesn't have a `map()` method (unlike `Object.entries()` which can be mapped)

ushkarev: we get [iterator helpers](https://github.com/tc39/proposal-iterator-helpers) in node 22; which goes LTS next week 💪

server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -25,20 +25,34 @@ declare module 'hmpo-form-wizard' {
     /** Use in complex generic forms that might allow multiple values in any fields; have checkbox components */
     type MultiValues = Record<string, MultiValue>
 
+    /** Operator function, returns true if submitted values matches condition */
+    type OperatorFunction = (
+      submittedValue: Value | MultiValue,
+      req: Request,
+      res: Express.Response,
+      condition: unknown,
```
ushkarev, line 33: (for future reference: i _think_ condition is `NextStep` or `NextStepCondition` but haven't confirmed it)

ushkarev (approved):

PR #251, opened 2024-10-31
Update dashboard

jonesj2193:
Updating the dashboard to include more features as designed in the prototype. 

Work so far:

- Added further filters
- Changed table columns to match prototype
- Made table sortable which works with filters and pagination.


server/routes/debug.ts
```diff
@@ -57,13 +119,17 @@ export default function makeDebugRoutes(services: Services): Record<string, Requ
...
         prisonId,
-        eventDateFrom: fromDate,
-        eventDateUntil: toDate,
+        incidentDateFrom: fromDate,
+        incidentDateUntil: toDate,
+        type: incidentType,
+        status: incidentStatuses,
         page: pageNumber - 1,
-        // sort: ['eventDateAndTime,ASC'],
+        size: 5,
```
ushkarev, line 131: i would suggest linking to `server/data/incidentReportingApi.ts:defaultPageSize` so that we use a single number throughout

server/routes/debug.ts
```diff
@@ -76,35 +142,87 @@ export default function makeDebugRoutes(services: Services): Record<string, Requ
...
-        urlPrefix,
-        'moj',
-        eventsResponse.totalElements,
-        eventsResponse.size,
-      )
-      const noFiltersSupplied = Boolean(!prisonId && !fromDate && !toDate)
 
-      const events = eventsResponse.content
-      const usernames = events.map(event => event.modifiedBy)
+      const noFiltersSupplied = Boolean(!prisonId && !fromDate && !toDate && !incidentType && !incidentStatuses)
```
ushkarev, line 161: should there be `reportingOfficer` in here too?

server/routes/debug.ts
```diff
@@ -76,35 +142,87 @@ export default function makeDebugRoutes(services: Services): Record<string, Requ
...
+
+      const reports = reportsResponse.content
+      const usernames = reports.map(report => report.reportedBy)
       const usersLookup = await userService.getUsers(res.locals.systemToken, usernames)
       const prisonsLookup = await prisonApi.getPrisons()
+      const reportingOfficers = Object.values(usersLookup).map(user => ({
```
ushkarev, line 167: i think most of the `Object.values(x)` can simply be `x` because they are arrays. you only need `Object.values` when you are iterating over an object.

ushkarev (commented):
this looks great! i would suggest adding a test or 2. i know that filtering is all a bit uncertain at the moment, but if we were to ignore filtering entirely in tests, we could add:
- it loads and displays _some_ reports
- it converts a username into a human name. if it is not found, the table simply displays the username

server/routes/debug.ts
jonesj2193: I will remove this, accidentally left that in from when I was changing it for testing purposes

server/routes/debug.ts
jonesj2193: It will be, I just hadn't put it in yet as it's not an active filter as such yet, as it doesn't yet do anything when you select it.

server/routes/index.ts
```diff
@@ -28,9 +29,9 @@ export default function routes(services: Services): Router {
 
   // view-only debug pages
   const debugRoutes = makeDebugRoutes(services)
-  get('/incidents', debugRoutes.eventList)
   get('/incidents/:id', debugRoutes.eventDetails)
   get('/reports/:id', debugRoutes.reportDetails)
+  router.use('/incidents', dashboard(services))
```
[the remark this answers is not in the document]
ushkarev: though i wonder if `get('/reports/:id', …)` might start to clash and we will need some more nesting of routers. best to keep your new one below the existing one as it is now

server/views/pages/debug/eventList.njk
```diff
@@ -5,14 +5,24 @@
...
 {% set pageTitle = applicationName + " - Home" %}
 
 {% block content %}
+  <h1>Incident reports</h1>
+  <div class="govuk-button-group">
+    {{ govukButton({
+    text: "Report an incident",
+    preventDoubleClick: true,
+    href: "/create-report",
+    isStartButton: true
```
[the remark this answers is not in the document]
jonesj2193: No good point, I forgot about the design of this, so have not mentioned it yet. But will add it to the list on the ticket

server/routes/dashboard.ts
```diff
@@ -0,0 +1,239 @@
...
+    const paginationParams = pagination(
+      pageNumber,
+      reportsResponse.totalPages,
+      urlPrefix,
+      'moj',
+      reportsResponse.totalElements,
+      reportsResponse.size,
+    )
+
+    res.render('pages/debug/eventList', {
```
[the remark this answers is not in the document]
jonesj2193: I meant to rename it already, I moved it and renamed it to dashboard to match the ts file. But can change it to reportsList if you think that is still a better name?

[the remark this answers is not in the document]
ushkarev: don't really mind. if the team call it dashboard, that's equally fine!

ushkarev (approved):

PR #252, opened 2024-10-31
IR-519: Automatically group contiguous questions into a page

xoen:
It is possible to group together questions when there is no branching, e.g. one or few questions lead to the next regardless of the answers. This adds that logic.

Of course some of the questions have a lot of possible answers and combine them could lead to pages with a lot of noise on screen.
To prevent this there is some logic to prevent any page to have more than a total number of 20 answers. This is a good rule of thumb and the results seem to be satisfactory: Questions with a lot of answers may not be grouped and questions with yes/no or few answers are naturally grouped together until the limit is reached.

The logic partitions the steps, so that each question only belongs to a single step.
This means that questions with multiple parents (e.g. where branching rejoin) are not grouped into any previous question (but they can be the first question of a group).
The rational for this is that linear questions naturally tend to be on a related topic, while a question which has multiple parents tend to be the beginning of a different area of the questionnaire.

For example, here part of the FINDS questionnaire:

Screenshot 2024-10-31 at 17 55 57

The first 2 questions are grouped together, while the 3rd is not because it has 2 parents.
As you can see the first two questions are grouped and they are about tobacco found, while the 3rd which is not grouped and it is about something else (weapons in this case).

### About the implementation
The grouping is currently done on top of the steps generation. This means the steps are generated as before (one per page) and then the grouping happens on top of this.

The `groupSteps()` is where all the "magic" happens. The tree traversal logic is a bit complicated but it is split into a few nested helper functions which operate on the `steps` (which are modified/merged as it goes).

A few things to note:
- `stepsWithSingleParent` is the Set constructed with the step ids of the steps with a single parent. This is to prevent steps with multiple parents to be grouped, as mentioned above.
- `answersCounts` is a Map that keeps track of how many answers a step currently has. This is important as we want to be sure the total doesn't exceed `MAX_ANSWERS_PER_PAGE`.
- the `groupSteps()` then invokes `groupStepsStartingAt()` with the first question/step.
- `groupStepsStartingAt()` is where the tree traversal logic and grouping happens:
  - the function checks if the step is eligible to be grouped with the following step, that is
    - the step has a single child (no branching)
    - the step is the only parent of the child
    - if the step is not eligible for grouping then we recursively attempt to group its children
  - if the step is eligible then we also check if merging with the single child would result in the step having too many questions.
    - if the merged step would still have a relatively small number of answers the two are merged:
      - the child's `next` value is used as the current step's `next`: As the child would then be the last question on the page this would determine the next step (and the `next` property is where the branching logic is)
      - the child's fields are added to the current step: In practice this simply mean that the child question will be shown in the current step.
      - finally the total number of answers for the current step is updated and the child step is deleted


server/@types/hmpo-form-wizard/index.d.ts
```diff
@@ -54,7 +54,7 @@ declare module 'hmpo-form-wizard' {
             | 'all'
             | 'some'
           value: Value | MultiValue
-          next: string
+          next: string | null
```
ushkarev, line 57: how can a next step condition point nowhere?

ushkarev: oh you mean when this is the final step? i feel that we should come back to this… firstly, the condition should probably be excluded? or perhaps it should point at the report's overview page (once that is wired up)?

server/data/testData/FINDS.ts
```diff
@@ -0,0 +1,5351 @@
...
+    },
+    {
+      prisonerRole: 'SUSPECTED_INVOLVED',
+      onlyOneAllowed: false,
+      active: true,
+    },
+  ],
+} as const
+
+export const steps: FormWizard.Steps = {
```
ushkarev, line 4511: interesting chicken and egg… was this generated?

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -3,6 +3,8 @@ import FormWizard from 'hmpo-form-wizard'
 import type { AnswerConfiguration, IncidentTypeConfiguration, QuestionConfiguration } from './types'
 import QuestionsController from '../../controllers/wip/questionsController'
 
+const MAX_ANSWERS_PER_PAGE = 20
```
ushkarev, line 6: i'm astonished that eslint was not upset 🤣 (ps: i like it)

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -46,9 +48,168 @@ export function generateSteps(config: IncidentTypeConfiguration): FormWizard.Ste
...
+ */
+function groupSteps(steps: FormWizard.Steps) {
+  const answersCounts: Map<string, number> = new Map()
+  const stepsWithSingleParent: Set<string | null> = buildStepsWithSingleParent()
+
+  /**
+   * Get the current number of answers for a given stepId
+   */
+  function getAnswersCountForStep(stepId: string) {
+    if (answersCounts.get(stepId) === undefined) {
```
ushkarev, line 95: how about `answersCounts.has(stepId)`? did that cause one of the many strange js troubles?

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -46,9 +48,168 @@ export function generateSteps(config: IncidentTypeConfiguration): FormWizard.Ste
...
+   * Count the number of answers for a given stepId
+   */
+  function countStepAnswers(stepId: string): number {
+    const step = steps[`/${stepId}`]
+    if (step === undefined) {
+      return 0
+    }
+
+    let count = 0
+    for (const nextStepCondition of Object.values(step.next as unknown as [{ value: string[] }])) {
```
ushkarev, line 111: `step.next` is an array isn't it? so you don't need the `Object.values(…)` wrapper? not that it harms anything, it just threw me as i thought this was an array. (unless it's not lol)

```typescript
for (const nextStepCondition of (step.next as unknown as [{ value: string[] }])) { … }
```

server/@types/hmpo-form-wizard/index.d.ts
xoen: It is very similar to how in the config the next question id is null as it's the final question.

> in any case, it feels weird to state a condition that is checked and matches, but ends up being ignored
why tho? If you answer in a certain way it means you're done it was the final question. It doesn't have to be special from the steps perspective (although of course it would be handled differently)

(bearing in mind potentially the last question may lead to other pages so this may have to be tweaked but it may be done separately from the main steps generation or not)

ushkarev (commented):
i LOVE this

server/@types/hmpo-form-wizard/index.d.ts
ushkarev: i'm looking at it from the point of view of the form wizard library: why would someone get me to check a condition and tell me it leads nowhere?

i'm not asking for a change, just raising a conceptual incongruity haha

ushkarev: >last question may lead to other pages
that is probably best left up to the controller to decide in `getNextStep`

server/data/testData/FINDS.ts
xoen: yeah, this is the generated config that I have moved under testData so that it doesn't change for whatever reason affecting the test.

server/data/incidentTypeConfiguration/formWizard.ts
xoen: eslint should shut it :P

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -46,9 +48,168 @@ export function generateSteps(config: IncidentTypeConfiguration): FormWizard.Ste
...
+    for (const [stepId, count] of stepsParentCount.entries()) {
+      if (count === 1) {
+        result.add(stepId)
+      }
+    }
+
+    return result
+  }
+
+  const startStepId = steps['/'].next as string
```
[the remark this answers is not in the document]
xoen: yeah, essentially "nothing to do"

server/data/incidentTypeConfiguration/formWizard.ts
xoen: to be honest I had to fight with the JS collections so much that I almost gave up at some point...I think `has()` is the proper way to check if a Set has the element but I wouldn't be surprised if does something weird (I'm going to try, tests should either pass or fail, if they pass I'll replace 👌 )

server/data/incidentTypeConfiguration/formWizard.ts
xoen: Correct, `step.next` is the array of `NextStepCondition` so you are right, `Object.values` is somewhat redundant.
Between JS weirdness and TS randomly complaining about things I must have played it safe plus it made it stick more explicit it was an array (as we also have `.next` `string | null` confusingly).

Lemme play with it, TS seems still somewhat happy with this:

```TypeScript
    for (const nextStepCondition of step.next) {
      count += (nextStepCondition as { value: string[] }).value.length
    }
```

which I think reads a bit better possibly?

xoen: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/252/commits/996af61f1844f8fce08443e416b1f949a64f3387 better 👍? worse 👎? I think it's slightly better but I still think not a fan of how we have to cast.

I suppose the challenge for TS with this kind of union types like `A | B | C` is that it doesn't know which one would it be. In this case is the following...

```TypeScript
{
          /** field, op and value. op defaults to '===' */
          field: string
          op?:
            | OperatorFunction
            | '>'
            | '>='
            | '<'
            | '<='
            | '=='
            | '==='
            | '!='
            | 'before'
            | 'after'
            | 'in'
            | 'all'
            | 'some'
          value: Value | MultiValue
          next: string | null
        }
```

I suppose we could have another type for this that we use in the definition of `NextStepCondition` but I think it would be just overcomplicating things possibly (and it may still not be enough because of the `Value | MultiValue`).

In this specific instance we only care about the `value` property being a `string[]` so possibly this is a bit more explicit.

ushkarev: yeah totally. i mean in the absolute general case we'd have to do a whole bunch of `Array.isArray(step.next)` and `'value' in item && Array.isArray(item.value)`. this would make the code worse, but technically appease type checks

ushkarev (approved):

PR #258, opened 2024-11-04
IR-572: Add basic role-checking to existing screens in the app

ushkarev:


Mjwillis (approved):
LGTM - thanks for picking this up!

PR #261, opened 2024-11-05
IR-419: Import NOMIS codes for all constants from api

ushkarev:
Depends on [api#201](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/201)


scripts/importDpsConstants.ts
```diff
@@ -61,9 +66,12 @@ main()
...
 function main() {
   const { scriptName, filePath, template } = parseArgs()
-  const { method, identifier, documentation, asEnum, includeNomisType } = template
+  const { method, identifier, documentation, asEnum } = template
 
-  const constants: (Constant | TypeConstant)[] = JSON.parse(fs.readFileSync(filePath, { encoding: 'utf8' }))
+  const constants: (Constant | TypeConstant | StaffRoleConstant | PrisonerRoleConstant | PrisonerOutcomeConstant)[] =
+    JSON.parse(fs.readFileSync(filePath, { encoding: 'utf8' }))
+  const includeNomisCode = constants.every(constant => 'nomisCode' in constant)
+  const includeNomisCodes = constants.every(constant => 'nomisCodes' in constant)
```
ushkarev, line 74: lets the api responses determine whether there are NOMIS codes

server/reportConfiguration/constants/correctionRequestReasons.ts
```diff
@@ -1,4 +1,4 @@
-// Generated with ./scripts/importDpsConstants.ts at 2024-10-23T09:57:41.903Z
```
xoen, line 1: I usually don't add stuff that didn't actually change in the commits.

For example in this case the content didn't actually change (aside from timestamp which is metadata really) - so the updated timestamp is a bit misleading.

Plus having these extra no-change commits may be a bit in our way if we ever need to dig for the history of a change.

ushkarev: i feel the opposite. if i see an older timestamp in a file, i would think it hasn't been recently regenerated so might have outdated content lol

xoen (approved):

PR #266, opened 2024-11-12
IR-631: Rename `prisonId` fields to `location` in reports and events

ushkarev:
…to match api changes with respect to creating, retrieving and updating their models

Depends on [api#206](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/206)


Mjwillis (approved):
Looks fine, any implication for syscon?

ushkarev:
@Mjwillis  this is the ui side, so none at all

PR #269, opened 2024-11-14
IR-576: View report answers

xoen:
Report questions page now display a report's responses to questions.

- moved questions page from `/:reportType/questions` to `/reports/:id/questions` as it now loads a report and shows the questions based on a given report's type
- the form's values are set from each of the questions' responses, unless a value is already set on the form (e.g. because user submitted a different value)
- matching between responses and fields values is done by answers' code: the `response` field of the answer's response is the "code" of the answer (e.g. `'YES'`, not an answer ID, nor the answer's label, e.g. `'Yes'`)
- values have to be set differently depending on whether the question allows multiple choices (`string[]` for checkboxes)  or a single response (`string` for radio buttons).
- each response can additionally require a date and/or a comment, this is based off the report type config. The report config is also used to determine these fields names (as their name is in the form `${question.id}-${answer.id}-(date|comment)`
- the API also currently prefixes question IDs with `QID-0...`, this may be problematic for updating questions' answers but for reading we can easily strip this prefix, which is what this does.
- finally, because of the dynamic way we import modules with the report types configuration and because of the difference between how the tests are run (TS) vs. how the application runs (transpiled JS)

### DIFF
--- a/server/controllers/wip/questionsController.ts
+++ b/server/controllers/wip/questionsController.ts
@@ -50,7 +50,11 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
         const fieldName = this.questionIdFromCode(question.code)
         const questionConfig: QuestionConfiguration = reportConfig.questions[fieldName]
         if (questionConfig === undefined) {
-          logger.warn(new Error(`Question with code '${fieldName}' not found in ${report.type}'s configuration.`))
+          logger.warn(
+            new Error(
+              `Report '${report.id}': Question with code '${fieldName}' not found in ${report.type}'s configuration.`,
+            ),
+          )
           // eslint-disable-next-line no-continue
           continue
         }
@@ -69,7 +73,7 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
           if (answerConfig === undefined) {
             logger.warn(
               new Error(
-                `answer with code '${answerCode}' not found in ${report.type}'s question '${questionConfig.id}' configuration.`,
+                `Report '${report.id}': Answer with code '${answerCode}' not found in ${report.type}'s question '${questionConfig.id}' configuration.`,
               ),
             )
             // eslint-disable-next-line no-continue
--- a/server/controllers/wip/questionsController.ts
+++ b/server/controllers/wip/questionsController.ts
@@ -37,7 +37,8 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
   ) {
     return super.getValues(req, res, async (err, values) => {
       if (err) {
-        return callback(err)
+        callback(err)
+        return
       }
 
       const formValues = { ...values }
@@ -98,7 +99,7 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
         }
       }
 
-      return callback(null, formValues)
+      callback(null, formValues)
     })
   }
 
--- a/server/controllers/wip/questionsController.ts
+++ b/server/controllers/wip/questionsController.ts
@@ -3,7 +3,6 @@ import type express from 'express'
 import { BaseController } from '../index'
 import { ReportWithDetails } from '../../data/incidentReportingApi'
 import format from '../../utils/format'
-import { getIncidentTypeConfiguration } from '../../reportConfiguration/types'
 import {
   AnswerConfiguration,
   IncidentTypeConfiguration,
@@ -17,17 +16,7 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
   }
 
   middlewareLocals(): void {
-    this.use(this.lookupReport)
     super.middlewareLocals()
-  }
-
-  async lookupReport(req: FormWizard.Request, res: express.Response, next: express.NextFunction): Promise<void> {
-    const { incidentReportingApi } = res.locals.apis
-    const reportId = req.params.id
-
-    res.locals.report = await incidentReportingApi.getReportWithDetailsById(reportId)
-    res.locals.reportConfig = await getIncidentTypeConfiguration(res.locals.report.type)
-    next()
   }
 
   getValues(
--- a/server/controllers/wip/questionsController.ts
+++ b/server/controllers/wip/questionsController.ts
@@ -1,3 +1,10 @@
+import { FormWizard } from 'hmpo-form-wizard'
+import type express from 'express'
 import { BaseController } from '../index'
 
-export default class QuestionsController extends BaseController {}
+export default class QuestionsController extends BaseController {
+  getBackLink(_req: FormWizard.Request, _res: express.Response): string {
+    // TODO: Change to `/reports/` page once we have it
+    return '/incidents/'
+  }
+}
--- a/server/controllers/wip/questionsController.ts
+++ b/server/controllers/wip/questionsController.ts
@@ -1,6 +1,15 @@
 import { FormWizard } from 'hmpo-form-wizard'
 import type express from 'express'
 import { BaseController } from '../index'
+import { ReportWithDetails } from '../../data/incidentReportingApi'
+import format from '../../utils/format'
+import { getIncidentTypeConfiguration } from '../../reportConfiguration/types'
+import {
+  AnswerConfiguration,
+  IncidentTypeConfiguration,
+  QuestionConfiguration,
+} from '../../data/incidentTypeConfiguration/types'
+import logger from '../../../logger'
 
 export default class QuestionsController extends BaseController<FormWizard.MultiValues> {
   getBackLink(_req: FormWizard.Request, _res: express.Response): string {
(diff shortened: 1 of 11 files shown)


server/controllers/wip/questionsController.ts
```diff
@@ -1,3 +1,111 @@
...
+      const formValues = { ...values }
+
+      const report = res.locals.report as ReportWithDetails
+      const reportConfig = res.locals.reportConfig as IncidentTypeConfiguration
+
+      for (const question of report.questions) {
+        const fieldName = this.questionIdFromCode(question.code)
+        const questionConfig: QuestionConfiguration = reportConfig.questions[fieldName]
+        if (questionConfig === undefined) {
+          logger.warn(new Error(`Question with code '${fieldName}' not found in ${report.type}'s configuration.`))
```
ushkarev, line 52: can we log the report reference or id pls? and the other errors below

also, is it "just" a warning or is it an error?

server/controllers/wip/questionsController.ts
```diff
@@ -1,3 +1,111 @@
...
+  }
+
+  getValues(
+    req: FormWizard.Request<FormWizard.MultiValues, string>,
+    res: express.Response,
+    callback: FormWizard.Callback<FormWizard.MultiValues>,
+  ) {
+    return super.getValues(req, res, async (err, values) => {
+      if (err) {
+        return callback(err)
```
ushkarev, line 40: this might work perfectly in practice now, but do be aware that in _general_ it's not a great idea to return the return value of a callback function. that could be an unrelated promise if the callback is async.

safer to:
```typescript
callback()
return
```

this point comes from being stung in the past

server/routes/questions/router.ts
```diff
@@ -11,26 +11,35 @@ const router = express.Router({ mergeParams: true })
...
-    const config = await getIncidentTypeConfiguration(reportType)
+    const config = await getIncidentTypeConfiguration(report.type)
 
     const steps = generateSteps(config)
     const fields = generateFields(config)
 
-    return wizard(steps, fields, {
-      name: `${reportType}-questions`,
+    const wizardRouter = wizard(steps, fields, {
+      name: `${reportId}-questions`,
```
ushkarev, line 30: great! less session clashing!

server/routes/questionsRoute.test.ts
```diff
@@ -0,0 +1,216 @@
...
+  let agent: Agent
+
+  beforeEach(() => {
+    agent = request.agent(app)
+    incidentReportingApi.getReportById.mockResolvedValue(reportWithDetails as ReportBasic)
+    incidentReportingApi.getReportWithDetailsById.mockResolvedValue(reportWithDetails)
+  })
+
+  it('form is prefilled with report answers, including date', () => {
+    reportWithDetails.type = 'DEATH_OTHER'
```
ushkarev, line 45: danger! shared `const reportWithDetails` being modified! when we add more tests to this file, we might not recall that type and questions must be written in every test case.

safer to either copy it, or make the `reportWithDetails` be a let variable populated with `beforeEach`.

server/controllers/wip/questionsController.ts
```diff
@@ -1,3 +1,111 @@
...
+  middlewareLocals(): void {
+    this.use(this.lookupReport)
+    super.middlewareLocals()
+  }
+
+  async lookupReport(req: FormWizard.Request, res: express.Response, next: express.NextFunction): Promise<void> {
+    const { incidentReportingApi } = res.locals.apis
+    const reportId = req.params.id
+
+    res.locals.report = await incidentReportingApi.getReportWithDetailsById(reportId)
```
ushkarev, line 28: the wrapping router already called `.getReportById`… should it instead have called this function instead so we don't need 2 lookups?

server/routes/questionsRoute.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      .get(reportQuestionsUrl)
+      .redirects(1)
+      .expect(200)
+      .expect(res => {
+        expect(res.text).not.toContain('There is a problem')
+        expect(res.text).toContain('name="61279" type="radio" value="POLICE REFERRAL" checked')
+        expect(res.text).toContain('name="61280" type="radio" value="NO" checked')
+        expect(res.text).toContain('name="61281" type="radio" value="NO" checked')
+        expect(res.text).toContain('name="61282" type="radio" value="NO" checked')
+        expect(res.text).toContain('name="61283" type="radio" value="YES" checked')
```
ushkarev, line 213: all these assertions rely on the current implementation of GOV.UK components.

i have an idea how to get round that but it's not a small change so i will park the thought for now!

server/controllers/wip/questionsController.ts
xoen: > can we log the report reference or id pls? and the other errors below
Good idea.

I started as an error tbh and then changed to a warning (as it doesn't exactly breaks) but I can try and change it back.

server/controllers/wip/questionsController.ts
xoen: I'll change 👍

server/routes/questionsRoute.test.ts
xoen: I'll have a look but tbh I'm setting the state just within the given test so it should be fine.

server/routes/questionsRoute.test.ts
xoen: Yeah this is starting to be integration-ey but I thought it was good enough for now. I am not sure how to test just the value-reading controller logic tbh. But I thought there was some value in the having a route test that implicitly test that as well.

In practice I don't imagine the govuk-component markup to change that much anytime soon.

server/reportConfiguration/types/index.ts
```diff
@@ -1,10 +1,24 @@
...
 export function getIncidentTypeConfiguration(type: string): Promise<IncidentTypeConfiguration> {
-  return import(`./${type}.js`).then(module => module.default.default)
+  // Import works differently when running TS directly (e.g tests)
+  const ext = __filename.endsWith('.js') ? 'js' : 'ts'
+  const configPath = path.resolve(__dirname, `./${type}.${ext}`)
+
+  return import(configPath)
+    .then(module => {
+      if (ext === 'js') {
+        return module.default.default
```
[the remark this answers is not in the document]
xoen: That's JavaScript modules for you 😬

server/controllers/wip/questionsController.ts
xoen: Messages now include the report ID: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/269/commits/da8b69d5688c56245182a2b56952e3dc9d529cca

They're also logged as error, not just warning: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/269/commits/95640dade1117e236bef1a78c5ac2c70aaa03fe5

server/controllers/wip/questionsController.ts
xoen: 59d34e39f9d961185426b4d28537fba24b1aeca7

server/controllers/wip/questionsController.ts
xoen: I noticed there was already a middleware that was doing the report lookup so I've tweaked things a little to reuse that:
- `populateReport` middleware gets the report with Details and also the report config now: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/269/commits/1bbb40c9ea251e31d93e4c8ff3c5b7a0f2d5c783
- Removed logic to get report/config from questions route/controller: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/269/commits/2df7b7d8dbd9eed24fdec518300849ea226c27df

server/middleware/populateReport.test.ts
```diff
@@ -22,27 +23,43 @@ describe('report-loading middleware', () => {
...
+  it('forwards error if report config cannot be loaded', async () => {
+    report.type = 'UNKNOWN' as Type
+    fakeApi.get(`/incident-reports/${report.id}/with-details`).reply(200, report)
+
+    const req = { params: { id: report.id } } as unknown as Request
+    const res = { locals: { apis: { incidentReportingApi: new IncidentReportingApi('token') } } } as Response
+    const next: NextFunction = jest.fn()
+    await populateReport()(req, res, next)
+
+    expect(next).toHaveBeenCalledWith(expect.objectContaining({ code: 'MODULE_NOT_FOUND' }))
```
ushkarev, line 61: is the next function called with an `Error` instance? dunno how stable/universal "MODULE_NOT_FOUND" is… so maybe `expect.any(Error)` is good enough? does `expect.objectContaining({ code: 'MODULE_NOT_FOUND' })` feel a little brittle to you?

server/routes/questionsRoute.test.ts
ushkarev: if u don't want to create the report in a `beforeEach`, can u leave a comment next to the `reportWithDetails` declaration noting that _every_ test must set the type and questions?

server/middleware/populateReport.test.ts
xoen: Not different from the similar assertion in the other test tbh. `any(Error)` risks to catch possibly other things so I'd rather have the more specific assertion and if the error ever change we'll cross that bridge. IMHO of course.

ushkarev (approved):

PR #274, opened 2024-11-20
Added activeAgencies to info as part of health checks and environment variables

jonesj2193:
To aid tile placement on DPS homepage.


server/middleware/setUpHealthChecks.ts
```diff
@@ -32,6 +32,7 @@ export default function setUpHealthChecks(applicationInfo: ApplicationInfo): Rou
         name: applicationInfo.applicationName,
       },
       productId: applicationInfo.productId,
+      activeAgencies: [process.env.SERVICE_ACTIVE_PRISONS],
```
xoen, line 35: What is the response's `activeAgencies` array meant to contain?

Is the environment variable a comma separated list of prison IDs? (or space separated?) Is this array meant to have these prison ids as elements or the whole string as coming from the environment variable?

e.g. Is `activeAgencies` expected to be something like:

```JSON
"activeAgencies": ["MDI", "LEI"] 
```

or

```JSON
"activeAgencies": ["MDI,LEI"] 
```

The former makes more sense but it is not what currently happens, but then again, I don't know what the client expects to find in this field (an array with a string or an array with the prison IDs, one per element).

xoen: In other words, perhaps this should actually split the environment variable by comma when building the `activeAgencies` array?

jonesj2193: Yeah this was something I was unsure of at the time with the use, but then Igor raised it and we realised so I have fixed it now 👍

xoen (approved):

PR #278, opened 2024-11-25
Update view report page

jonesj2193:
Making changes and adding unit tests to move the view report/report summary page closer to final design. 

Note: There are links put in place on the page that will not work yet. They are just placeholders to replace with the appropriate links once the relevant work has been done. 

This PR will NOT deliver the page's final design.


server/views/pages/debug/reportDetails.njk
```diff
@@ -2,138 +2,246 @@
...
-        text: "Reported by"
+    },
+    {
+      key: {
+        text: "Description"
       },
-      {
-        text: usersLookup[report.reportedBy].name or report.reportedBy
+      value: {
+        html: report.description
```
ushkarev, line 59: it's safer to have `text: x` in these items!
(so that something like `<script>alert('lol')</script>` in the description doesn't rickroll users 😉)

server/views/pages/debug/reportDetails.njk
```diff
@@ -2,138 +2,246 @@
...
+        {% if prisonerInvolved.outcome %}
+          {%- set outcome = prisonerOutcomeLookup[prisonerInvolved.outcome] or prisonerInvolved.outcome -%}
+        {% else %}
+          {%- set outcome = 'No outcome' -%}
+        {% endif %}
+
+        {%- set prisonerURL = dpsUrl + '/prisoner/' + prisonerInvolved.prisonerNumber -%}
+        {%- set prisonerRows = (prisonerRows.push({
+          key: {
+            html: '<a href=' + prisonerURL + ' class="govuk-link">' + prisonerName + '</a>'
```
ushkarev, line 113: this block would be a little better to do in a slightly better way. for instance, `prisonerName` is not escaped and the href attribute won't have quotes. you could write the whole html block outside this with:

```nunjucks
{% set prisonerURLWhatever %}
  <a href="{{ prisonerURL }}">{{ prisonerName }}</a>
{% endset %}
```
so that this last line becomes `html: prisonerURLWhatever`

server/views/pages/debug/reportDetails.njk
ushkarev: there are a few more `html: x` below that would also be safer as `text: x`

server/routes/viewReport.test.ts
```diff
@@ -0,0 +1,227 @@
...
+        expect(res.text).toContain('A2222BB')
+        expect(res.text).toContain('Suspected involved - No outcome')
+      })
+  })
+  it('should render staff involved with roles', () => {
+    return request(app)
+      .get('/reports/6543')
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('lev79n')
```
ushkarev, line 111: ah this should turn into a staff member's name! you should add a few more entries to the `const users: Record<string, User> = …` by referencing the fake users in /server/data/testData/prisonApi.ts

ushkarev (commented):
looking good! a few suggestions…

server/routes/viewReport.test.ts
jonesj2193: Have included this now. I just took the name and username from the test data staff member. Did go to import it but it is a different structure to what is returned by the `getUsers` function. So I just manually added it to the mock result instead.

ushkarev (approved):

PR #289, opened 2024-12-09
IR-685: Add special exceptions to sentence case conversion

ushkarev:
…where word boundaries prevent using the acronym exception list. This is a minor tweak to #287 which slightly improves conversions from NOMIS-speak.


xoen (dismissed):

server/utils/utils.ts
```diff
@@ -354,10 +354,14 @@ function postProcessSentenceCase(input: string): string {
...
     },
+    {
+      match: /\bCategory ([ABCDE])\b/i,
+      replace: (_: string, category: string) => `Category ${category.toUpperCase()}`,
+    },
   ]
   let output = input
   for (const { match, replace } of exceptions) {
-    output = output.replace(match, replace)
+    output = output.replace(match, replace as string)
```
ushkarev, line 364: ts can't figure out that both options of the `replace` union in the exceptions above are allowed so pretending that it's only ever string :(

xoen (approved):

PR #290, opened 2024-12-11
Hide debug info in questions page

xoen:
It is still easily possible to show the debug info by temporarily toggle the `if` statement condition in local environment when needed.

We can remove later on, but for now it's useful to be able to quickly have a look at these information.


ushkarev (approved):
still think that `{% if req.query.debug %}` would've been more flexible. otherwise u need to PR turning it on and if it's always off, not even much need for it to be checked in

PR #291, opened 2024-12-11
Report view: Questions case: Account for `QID-` prefix

xoen:
When displaying questions/answers in the report view strip any possible `QID-` prefix that may still be present in the data returned by the API.

Long term, once all the data will no longer have
these prefixes (e.g. all environment have been re-migrated) we can remove this simple logic. Although for codes without this prefix this is a no-op.

### DIFF
--- a/server/data/incidentTypeConfiguration/types.ts
+++ b/server/data/incidentTypeConfiguration/types.ts
@@ -57,3 +57,13 @@ interface PrisonerRoleConfiguration {
 export function findAnswerConfigByCode(answerCode: string, questionConfig: QuestionConfiguration): AnswerConfiguration {
   return questionConfig.answers.find(answerConfig => answerConfig.code.trim() === answerCode.trim())
 }
+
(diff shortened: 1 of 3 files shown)


server/data/incidentTypeConfiguration/types.ts
```diff
@@ -57,3 +57,13 @@ interface PrisonerRoleConfiguration {
 export function findAnswerConfigByCode(answerCode: string, questionConfig: QuestionConfiguration): AnswerConfiguration {
   return questionConfig.answers.find(answerConfig => answerConfig.code.trim() === answerCode.trim())
 }
+
+/** Strips `QID-0...` prefix and returns the question ID
+ *
+ * TODO: Remove QID-stripping logic once removed from API
+ */
+export function stripQidPrefix(code: string): string {
```
xoen, line 65: Moved here from `QuestionsController` to also use it in `viewReport` route.

PR #293, opened 2024-12-13
IR-722: Require report involvements to have names

ushkarev:
Adapting to changes in [api#237](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/237)


server/controllers/addPrisoner/addPrisoner.ts
```diff
@@ -64,6 +64,8 @@ export default class AddPrisoner extends FormInitialStep {
 
       const newPrisonerData: PrisonerInvolvement = {
         prisonerNumber: res.locals.prisoner.prisonerNumber as string,
+        firstName: 'Not',
+        lastName: 'Known',
```
ushkarev, line 68: i'd say this doesn't matter since the whole add-a-prisoner process has not been designed so it's not worth the effort wiring this up

xoen (approved):

PR #300, opened 2024-12-20
IR-738: Remove reasons for correction requests

ushkarev:
This field is not available in NOMIS data and does not appear necessary given progressing designs

See also: [api#254](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/254)


jonesj2193 (approved):
LGTM

PR #301, opened 2024-12-20
IR-732: Delete now-irrelevant questions when answering questions

xoen:
When a user answer/update questions some of the existing responses may no longer make sense, so they should be deleted.

For example, let's say the user previously answered as follows:
Q1: Is the location known? YES -> Q2 (NO would go to Q3)
Q2: Location? GYM -> Q3
Q3: Any serious injuries? YES
...

Now if the user goes back and change the answer to the first question to NO, of course the answer to Q2 wouldn't make any sense, so it should be deleted.

The logic is in the `QuestionsToDelete` service, probably a good start to review this PR is to start with the tests to this service and follow the tests using the linked SVG questionnaires charts.

The original aim of the ticket was to do the simplest sensible thing, which is delete all subsequent questions. However this may not be as simple as it sound because of the way we store questions in dictionaries.
However, as a result, the logic to determine what to delete is traversing the questionnaire tree which means we can actually do a better job at retaining existing questions as much as _possible_. This is an attractive property because we avoid forcing users to re-answers in many cases.

### Key idea
The fundamental idea behind the way this works is that we use the answered questions to traverse the questionnaire tree.
We know which questions were answered before the update - so we can delete any questions which were not visited during this traversal.

The logic does something slightly clever, as sometimes a user may have not answered a given question yet but as there wouldn't be any branching we can continue with the traversal.

The important consequence of this is that when a user changes an answer to a question which branches we can follow the tree and potentially re-join the main flow...this means the questions after the branch rejoins can be retained (as we know given the current answers they'll be reached anyway).

Of course this is not always possible, because sometimes questionnaires's branches have sub-branches (and the traversal can follow every possible combination, and we can't predict how a user may answer all the questions) so in some cases future questions/answers **will** be deleted. But in many cases the user existing responses can be preserved.

This is probably tricky explain with words, but probably **look at the charts is much easier**.

Screenshot 2024-12-20 at 13 24 12

Here for example:
1. user went back and changed the answer to "Were any drugs found?" to YES (was NO).
2. previously they said "No" and also answered questions after the branch ("Was drug equipment found?" and "Describe  drug equipment found")
3. the questions in the branch are not answered yet but there is no branching
4. because of 3) the questions in the branch can be traversed until it reaches previously answered questions (2)
5. as a result, previously answered questions (2) are reachable and **retained** (and user is happy as they don't need to re-answer them)

Hope this makes sense, tests have other cases like this that may help understand how this work. And as usual, if unclear I'm happy to have a chat :)


server/services/questionsToDelete.ts
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * The logic to determine questions to delete only needs
+ * the question `code` and the responses' `response`
+ */
+interface AnsweredQuestion {
+  code: string
+  responses: { response: string }[]
+}
+
+export default class QuestionsToDelete {
```
ushkarev, line 14: not really a big deal, but why a class with only static methods? there's no share state?

server/routes/questions/router.test.ts
```diff
@@ -328,6 +331,24 @@ describe(`Submitting questions' responses`, () => {
...
 
+    const questionsResponse: Question[] = [
+      {
+        code: '45054',
+        question: 'WERE THE POLICE INFORMED OF THE INCIDENT',
+        additionalInformation: null,
+        responses: [
+          {
+            response: 'YES',
+            responseDate: new Date(),
```
ushkarev, line 342: technically `responseDate` is `parseDateInput(responseDate)`

server/routes/questions/router.test.ts
```diff
@@ -372,6 +394,31 @@ describe(`Submitting questions' responses`, () => {
...
 
+    const questionsResponse: Question[] = [
+      {
+        code: '67179',
+        question: 'DESCRIBE HOW THE ITEM WAS FOUND (SELECT ALL THAT APPLY)',
+        additionalInformation: null,
+        responses: [
+          {
+            response: 'BOSS CHAIR',
+            responseDate: new Date(),
```
ushkarev, line 405: technically `responseDate` is null (and for next response too)

server/routes/questions/router.test.ts
```diff
@@ -467,6 +515,113 @@ describe(`Submitting questions' responses`, () => {
       },
     ]
 
+    const questionsResponse: Question[] = [
```
ushkarev, line 518: technically, `responseDate` in all responses in this list is null

server/services/questionsToDelete.ts
```diff
@@ -0,0 +1,115 @@
...
+    start: string,
+    answeredQuestions: AnsweredQuestion[],
+  ): string | null {
+    const answeredQuestion = answeredQuestions.find(question => question.code === start)
+    let questionConfig = null
+    let answerConfig = null
+
+    if (answeredQuestion) {
+      questionConfig = config.questions[answeredQuestion.code]
+      if (questionConfig) {
```
ushkarev, line 91: if `questionConfig` is **not** found for an answered question, this answer essentially becomes unvisited, right? (and therefore would get deleted)

ushkarev (approved):

server/services/questionsToDelete.ts
xoen: At some point I almost had the config as state but I removed. I still prefer this to the alternative (a bunch of functions) as these methods are all related to each other.

server/services/questionsToDelete.ts
xoen: This is a bit of an edge case but I covered it in case for whatever reason the report had some questions which we cannot find in the config.

It would then try the else branch, not find the question config there neither and eventually leave the if/else and return null as start can't be visited (so yes, it would ultimately delete the questions with id `start`)

server/routes/questions/router.test.ts
xoen: This is to mock the return value of `addOrUpdateQuestionsWithResponses()` which returns `Question[]` (well, a `Promise<Question[]` to be precise). So responseDate is actually a `Date`, unless I'm missing something 🤔

server/services/questionsToDelete.ts
ushkarev: well a file with 1 public and 2 private methods achieves the same these-are-related aim. no, i'm not demanding a change here ;)
but i was surprised to see that the `this` keyword works in a static context

server/routes/questions/router.test.ts
xoen: True, these answers have `commentRequired: false` (although in practice it doesn't make a difference for the test)

server/routes/questions/router.test.ts
xoen: Fair enough. I'll change

(I was more focused on the overarching flow of questions/answer rather then the comments/date which don't make much of a difference in this context...so there was a fair amount of copy and paste)

server/services/questionsToDelete.ts
ushkarev: there is no else branch to no `questionConfig`

server/routes/questions/router.test.ts
ushkarev: it totally doesn't make any practical difference. but realistic test/mock data structures make it more readable and easier to reason about… otherwise i might look in 2 months time and wonder where the date came from

xoen: Changed test data to have `null` dates when they're not required in this commit: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/301/commits/e2c1965538098f906eb63c448ba85c00e6adda0b

server/routes/questions/router.test.ts
xoen: Changed test data to have `null` dates when they're not required in this commit: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/301/commits/e2c1965538098f906eb63c448ba85c00e6adda0b

server/services/questionsToDelete.ts
xoen: Yes it does work. Sorry you are right, I didn't look at the code properly when answering you, so yeah it does not enter the `else` branch of course.

However notice how critically I'm declaring the `questionConfig` and `answerConfig` variables before the the branch and they're `null` unless we can correctly find the config/answer and unambiguously determine the next question to visit.

So at the end of the day, if we can't find the question config or the question was not answered and there is no branching (or if it's the last question!) this method would return null and end end the traversal. The end result is, as you said, that this unknown question would not be visited and it will be deleted.

Hope this makes sense and it's less confusing this time 😅

server/services/questionsToDelete.ts
xoen: We shouldn't be afraid of using classes to group things ;) - In my mind determine which questions to delete is providing that specific service and the traversal, etc. stuff is internal implementation that the client doesn't care/need to know ☺️

server/services/questionsToDelete.ts
ushkarev: great, that's what i thought. i like that self-healing aspect of this cos it's always someone feasible for a rando-question to appear in the answer set… a weird sync error; some data race; technically-erroneous-according-to-config nomis data…

server/routes/questions/router.test.ts
ushkarev: yes it's a date, but technically not “now” ;)

ushkarev (approved):

PR #305, opened 2025-01-14
IR-743 dashboard search

jonesj2193:
Dashboard search functionality added as per IR-743.

In this branch:

Search added to dashboard
- Search will look for matches on incident reference number or prisoner number involved fields. Search will do either or depending on regex match on the pattern of the submitted value.
- Search input has validation that if user has submitted a value that is neither only digits or matches the prisoner id pattern, an error message will appear and the search will not be filtered.

Unit tests have been added to check this functionality.

### DIFF
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -106,12 +106,14 @@ export type ReportWithDetails = ReportBasic & {
 }
 
 export type GetEventsParams = {
+  reference: string
   location: string
   eventDateFrom: Date // Inclusive
   eventDateUntil: Date // Inclusive
 } & PaginationSortingParams
 
 export type GetReportsParams = {
+  reference: string
   location: string | string[]
   source: InformationSource
   status: Status
@@ -227,7 +229,8 @@ export class IncidentReportingApi extends RestClient {
   }
 
   async getEvents(
-    { location, eventDateFrom, eventDateUntil, page, size, sort }: Partial<GetEventsParams> = {
+    { reference, location, eventDateFrom, eventDateUntil, page, size, sort }: Partial<GetEventsParams> = {
+      reference: null,
       location: null,
       eventDateFrom: null,
       eventDateUntil: null,
@@ -240,6 +243,9 @@ export class IncidentReportingApi extends RestClient {
       page,
       size,
       sort,
+    }
+    if (reference) {
+      query.reference = reference
     }
     if (location) {
       query.location = location
@@ -277,6 +283,7 @@ export class IncidentReportingApi extends RestClient {
 
   async getReports(
     {
+      reference,
       location,
       source,
       status,
@@ -292,6 +299,7 @@ export class IncidentReportingApi extends RestClient {
       size,
       sort,
     }: Partial<GetReportsParams> = {
+      reference: null,
       location: null,
       source: null,
       status: null,
@@ -312,6 +320,9 @@ export class IncidentReportingApi extends RestClient {
       page,
       size,
       sort,
+    }
+    if (reference) {
+      query.reference = reference
     }
     if (location) {
       query.location = location
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -81,7 +81,7 @@ export type ReportBasic = {
   reportReference: string
   type: Type
   incidentDateAndTime: Date
-  location: string
+  location: string | string[]
   title: string
   description: string
   reportedBy: string
@@ -112,7 +112,7 @@ export type GetEventsParams = {
 } & PaginationSortingParams
 
 export type GetReportsParams = {
-  location: string
+  location: string | string[]
   source: InformationSource
   status: Status
   type: Type
--- a/server/routes/dashboard.test.ts
+++ b/server/routes/dashboard.test.ts
@@ -1,30 +1,25 @@
 import type { Express } from 'express'
 import request from 'supertest'
 
-import { PrisonApi } from '../data/prisonApi'
 import { appWithAllRoutes } from './testutils/appSetup'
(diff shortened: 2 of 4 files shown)


server/data/incidentReportingApi.ts
```diff
@@ -81,7 +81,7 @@ export type ReportBasic = {
   reportReference: string
   type: Type
   incidentDateAndTime: Date
-  location: string
+  location: string | string[]
```
ushkarev, line 84: this can’t be an array, we only store one location (prison id) in the db. i figure you meant to put this inside `GetEventsParams` – oops i forgot to update the types in this file after i changed the api!

jonesj2193: Ah yeah I put it in the wrong place, was meant to be in the `GetEventParams` along with the `GetReportParams`

server/routes/dashboard.ts
```diff
@@ -115,6 +123,23 @@ export default function dashboard(service: Services): Router {
       toDate = null
       errors.push({ href: '#toDate', text: `Enter a valid to date, for example ${todayAsShortDate}` })
     }
+    let prisonerId: string = null
+    let referenceNumber: string = null
+    if (searchID) {
+      // Test if search is for a prisoner ID and use if so
+      if (searchID.match(/^[a-zA-Z][0-9]{4}[a-zA-Z]{2}$/)) {
```
ushkarev, line 130: i'd recommend trimming whitespace around `searchId` so that "A1401AE " still works

server/views/pages/dashboard.njk
```diff
@@ -20,24 +20,26 @@
     }) }}
   </div>
   <div class="horizontal-form__wrapper govuk-!-padding-3 govuk-!-margin-bottom-5">
-    <h2 class="govuk-heading-m">Filter by</h2>
+    <h2 class="govuk-heading-l">Incident reports</h2>
```
ushkarev, line 23: the first heading should still be an h1 (even if it’s supposed to look small) for proper doc structure. e.g. it helps screen readers.

server/routes/dashboard.ts
```diff
@@ -123,19 +148,32 @@ export default function dashboard(service: Services): Router {
...
+
+    // Set locations to user's caseload
+    let searchLocations: string[] | string = userCaseloadIds
+    // Overwrite locations with chosen filter if it exists
+    if (location) {
+      searchLocations = location
+    }
+
     // Get reports from API
     const reportsResponse = await incidentReportingApi.getReports({
```
ushkarev, line 160: you may wish to do this as a separate pr, but i think the api should not be called at all if `errors` is not empty

server/routes/dashboard.test.ts
```diff
@@ -44,12 +39,17 @@ describe('GET dashboard', () => {
...
     userService.getUsers.mockResolvedValueOnce(users)
 
-    const prisons = {
-      LEI: leeds,
-      MDI: moorland,
+    const expectedParams: Partial<GetReportsParams> = {
+      location: ['MDI'],
+      incidentDateFrom: undefined,
+      incidentDateUntil: undefined,
+      involvingPrisonerNumber: null,
```
ushkarev, line 46: is it significant that one is undefined and one is null? i'm supposing that a `null` is actually sent to the api? presumably that still works since the api considers all parameters as optional here.

if that’s what’s happening here, ok, but one thing to note is the api _won’t_ treat undefined and null as the same for updates to nullable fields (e.g. prisoner involvement outcome) cos it needs a way to both 1) remove an existing value (by sending null) and 2) keep the existing value (by not sending anything; ie. undefined)

ushkarev: long story short: might be a bit safer and more consistent to stick with undefineds. probably maybe

(i mean at the point where you _call_ the api class endpoint, not this test matcher)

server/routes/dashboard.test.ts
```diff
@@ -93,4 +87,162 @@ describe('GET dashboard', () => {
...
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).not.toContain(
+          'Enter a valid incident reference number or offender ID. For example, 12345678 or A0011BB',
+        )
+        expect(res.text).toContain('Clear')
+        expect(incidentReportingApi.getReports).toHaveBeenCalledWith(expectedParams)
+      })
+  })
 })
```
ushkarev, line 248: could you add a test pls with all fields filled in with a distinct value? that way we can be assured that each one is propagated to a parameter in the api call.

ushkarev: there are a fair few other test cases that i would say are necessary to do… but i suppose they can come later since this pr is mainly focused on prisoner number / ref search box. maybe we should add a ticket to make sure we come back to it. i'm thinking things like sorting/pagination but there are likely others… e.g. are fields properly remembered when you go to page 2?

server/routes/dashboard.test.ts
jonesj2193: Yeah I noticed when putting the unit tests together about this and thought it didn't seem right but was still working I guess. But I have changed it to be more consistent now

server/routes/dashboard.ts
jonesj2193: I think I will do this as part of the display/visuals ticket.

server/routes/dashboard.test.ts
jonesj2193: There are a bunch of tests that will be needed altogether. I was going to add tests per filter as I work on them over the next couple of tickets, and will then cover everything remaining as part of the visuals/display ticket.

jonesj2193: I have added in a unit test with all filter query parameters in to check they get submitted correctly.

ushkarev (approved):

PR #312, opened 2025-01-21
IR-68: SCSS tweak

ushkarev:
The sass compiler issues a deprecation warnings about ordering of nested declarations but only in the fallback header component. Reordering plain properties first appeases it, but doesn’t change the actual effect on layout. (Users never see this fallback header anyway)


Mjwillis (approved):
LGTM

PR #315, opened 2025-01-22
IR-740: Report question display

ushkarev:
The key change in this PR is the addition of a utility that walks through the questions in a report indicating whether they are completed and what page they appear on.

This has enabled numbering questions and the pages they appear on.

NB: question _editing_ is buggy due to how form wizard works / is set up. Ie. going back to change a response will not always work.


Mjwillis (approved):
Nice !

PR #316, opened 2025-01-23
Added accessible autocomplete component.

jonesj2193:
Configured to be a usable macro in nunjucks templates. 
Applied component to establishment and incident type on dashboard


Mjwillis (approved):
LGTM

Mjwillis (dismissed):
Shouldn't there be integration tests for this?

assets/scss/application.scss
```diff
@@ -7,6 +7,8 @@ $govuk-page-width: $moj-page-width;
 
 @import 'govuk-frontend/dist/govuk';
 @import '@ministryofjustice/frontend/moj/all';
+@import '../../node_modules/accessible-autocomplete/dist/accessible-autocomplete.min';
```
ushkarev, line 10: i think the import path should be `'accessible-autocomplete/dist/accessible-autocomplete.min'` because the node modules folder is already included in the sass load paths and because a relative import is more brittle – it depends more strongly on where this file lives.

assets/scss/components/_horizontal-content.scss
```diff
@@ -70,4 +70,8 @@
 
 .horizontal-form__wrapper {
   background-color: govuk-colour("light-grey");
+  .autocomplete__input {
+    background-color: white;
+    width: 29em
```
ushkarev, line 75: we ought to not specify exact widths in general, but it’s fine for now since there will be a separate jobby to align all the boxes later

assets/scss/components/_horizontal-content.scss
```diff
@@ -70,4 +70,8 @@
 
 .horizontal-form__wrapper {
   background-color: govuk-colour("light-grey");
+  .autocomplete__input {
+    background-color: white;
```
ushkarev, line 74: we can use the govuk-frontend colour set to be more future proof by specifying `$govuk-body-background-colour`.

assets/scss/components/accessible-autocomplete.scss
```diff
@@ -0,0 +1,7 @@
+@import 'govuk-frontend/dist/govuk';
+
+.autocomplete__option {
+  @extend .govuk-body;
```
ushkarev, line 4: is this to get the font right?

server/views/macros/accessible-autocomplete/template.njk
```diff
@@ -0,0 +1,32 @@
...
+  <div id="{{ params.id }}-hint" class="govuk-hint">
+    {{ params.hint }}
+  </div>
+  {% endif %}
+  {% if params.errorMessage %}
+    <span id="{{ params.id }}-error" class="govuk-error-message">
+      <span class="govuk-visually-hidden">Error: </span>{{ params.errorMessage.text }}
+    </span>
+  {% endif %}
+  <select id="{{ params.id }}" name="{{ params.id }}"class="govuk-input govuk-input--width-30">
```
ushkarev, line 17: class attribute needs a space before it

server/views/pages/dashboard.njk
```diff
@@ -9,6 +9,7 @@
 {% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
 {% from "moj/components/date-picker/macro.njk" import mojDatePicker %}
 {% from "moj/components/pagination/macro.njk" import mojPagination %}
+{% from "../macros/accessible-autocomplete/macro.njk" import accessibleAutocomplete %}
```
ushkarev, line 12: (fyi: don't need the `../` at the beginning of the import)

assets/scss/components/_horizontal-content.scss
jonesj2193: Yeah I know Dan wants them to be 50% width, I just fixed it at this for now until I got back through and fix the stylings.

assets/scss/components/accessible-autocomplete.scss
jonesj2193: I think it was to inherit certain aspects correctly? I copied it from what they did in the jobs board ui, I haven't actually tested removing it, so may not be needed.

ushkarev (approved):

PR #320, opened 2025-01-29
IR 745 status filters

jonesj2193:
Work done to show all filters for DW view or Work list for RO viewing the dashboard.


assets/scss/components/_horizontal-content.scss
```diff
@@ -74,4 +74,14 @@
...
+  @media (min-width:40.0625em) {
+    .govuk-checkboxes--small {
+      display: flex;
+      flex-wrap: wrap;
+      align-items: flex-start
+    }
+    .govuk-checkboxes__label {
+      max-width: None;
+    }
+  }
```
ushkarev, line 86: i'd recommend a similar `background-color: $govuk-body-background-colour;` you used for autocomplete boxes on the appropriate element cos the checkbox appears grey

server/routes/dashboard.ts
```diff
@@ -1,24 +1,34 @@
...
+export type IncidentStatuses = Status | WorkList
 
 interface ListFormData {
   searchID?: string
   location?: string
   fromDate?: string
   toDate?: string
   incidentType?: Type
-  incidentStatuses?: Status
+  incidentStatuses?: IncidentStatuses
```
ushkarev, line 31: i can't totally tell because of the type assertions that are forcing type conversion below… but isn't this an _array_ or (real api) status codes or work list codes?

server/routes/dashboard.ts
```diff
@@ -159,14 +171,28 @@ export default function dashboard(service: Services): Router {
       searchLocations = location
     }
 
+    let searchStatuses: Status | Status[]
+    // Replace status mappings if RO viewing page
+    if (userRoles.includes(roleReadWrite) && !userRoles.includes(roleApproveReject)) {
+      if (typeof incidentStatuses === 'object') {
```
ushkarev, line 177: `Array.isArray(incidentStatuses)` is closer cos you use this var as an array on the line below

server/reportConfiguration/constants/statusMappings.ts
```diff
@@ -0,0 +1,34 @@
...
+export type WorkList = WorkListDetails['code']
+
+export type WorkListStatusMapping = { [key: WorkList]: Status[] }
+
+export const workListStatusMapping: WorkListStatusMapping = workListMapping.reduce(
+  (prev, workListOption) => ({ ...prev, [workListOption.code]: workListOption.statuses }),
+  {},
+)
+
+// { code: 'POST_INCIDENT_UPDATE', description: 'Post-incident update', nomisCode: 'PIU' },
```
ushkarev, line 33: would you mind pls putting `TODO:` -type comments so that it's easier to know that this is something we must return to? IDEs highlight them and intellij has a panel that looks across the whole project

server/routes/dashboard.ts
```diff
@@ -255,6 +292,7 @@ export default function dashboard(service: Services): Router {
       incidentTypes,
       statusItems,
       typesLookup,
+      checkboxLabel,
```
ushkarev, line 295: just a suggestion, but maybe this var name could be a little more specific? the label relates to status filters of different kinds.

assets/scss/components/_horizontal-content.scss
jonesj2193: So I wrestled with this for a while trying to make the checkbox background white but it wasn't working. An issue with the opacity, as it blocks it all out, but then if I reduce it, you just get a huge white square. So I think the whole thing may need to be re-written/designed as such. But as I was spending so much time on it, I left it to come back to when I do the visuals/style work as part of the next ticket.

server/routes/dashboard.ts
jonesj2193: So at this stage I created a new type to accept either direct status or work list values, so that they appear against the query. But later on, the work list options get converted into the list of corresponding statuses, that variable is set to be a Status or list of, and that is what is parsed into the api call.

server/reportConfiguration/constants/statusMappings.ts
jonesj2193: Ah yeah I did mean to come back and do something with this, I left them in as a reminder of the statuses not mapped. I have spoken to Dan about it. Will add the `TODO`

ushkarev (approved):

PR #322, opened 2025-01-30
IR-770: Package accessible autocomplete component into JS bundle

ushkarev:
This:
- turns the component into a proper govuk-frontend type of component so that it is initialised alongside everything else in the design systems
- simplifies component HTML template because it’s almost identical to `govukSelect`
- simplifies component-specific style overrides
- adds types for govuk-frontend’s new configurable component (which could be used later to configure the accessible autocomplete component)
- and also tidies up relative paths in nunjucks template imports and extensions


jonesj2193 (approved):
LGTM

PR #325, opened 2025-01-31
Ir 768 dashboard display

jonesj2193:
Various changes to display on the dashboard. Changes include:

- Columns displayed in table depend on number of establishments in caseload. More than 1 shows establishment column, if only 1 then column is replaced with "reported by". Unit tests added to test this. 
- Description column is limited to 2 rows and an ellipsis ends the shown text. 
- Filters have been arranged using rows and fixing the max widths as a portion of the display width, with minimum widths applied. 
- Checkboxes have been fixed to show white backgrounds.
- Moved dashboard into it's own subfolder for associated files.
- Some other minor wording tweaks to match design.


Mjwillis (dismissed):
LGTM

assets/scss/components/_horizontal-content.scss
```diff
@@ -70,9 +71,23 @@
...
+  .govuk-grid-column-one-quarter-from-desktop {
+    box-sizing: border-box;
+    padding: 0 15px 0 0;
+  }
   .autocomplete__input {
     background-color: $govuk-body-background-colour;
-    width: 29em
+    width: 100%
   }
   @media (min-width:40.0625em) {
```
ushkarev, line 92: how did you decide on `40.0625em`? is this supposed to be desktop size?

assets/scss/components/_horizontal-content.scss
```diff
@@ -32,8 +32,6 @@
 
 .horizontal-form {
```
ushkarev, line 33: the way you're using rows now means there’s nothing left of `.horizontal-form` that has an effect, right? it looks as though there will be clashing or overriding styles…

ushkarev: actually on that point – did you & Dan deliberately want the presence of error messages to no longer change the layout to vertical?

server/views/pages/dashboard.njk
```diff
@@ -44,88 +46,92 @@
...
-          legend: {
-            text: statusCheckboxLabel,
-            classes: "govuk-fieldset__legend--s"
-          }
-        },
-        items: statusItems | govukMultipleCheckedItems(formValues.incidentStatuses),
-        errorMessage: errors | findFieldInGovukErrorSummary("incidentStatuses")
-      }) }}
+      <div class="govuk-grid-row">
+        {{ govukCheckboxes({
```
ushkarev, line 120: if you have a `div.govuk-grid-row` you need to nest a `div.govuk-grid-column-?` to make the alignment work as intended

server/views/pages/dashboard.njk
```diff
@@ -1,4 +1,4 @@
-{% extends "partials/layout.njk" %}
+{% extends "../partials/layout.njk" %}
```
ushkarev, line 1: i think it's easier when template imports and extensions use paths from the root cos then files can be moved around or copied without changing these relative paths

server/routes/dashboard/dashboard.test.ts
```diff
@@ -275,8 +275,80 @@ describe('GET dashboard', () => {
...
+      .get('/reports')
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('>Reference number <')
+        expect(res.text).toContain('>Type <')
+        expect(res.text).toContain('>Incident date <')
+        expect(res.text).toContain('>Description<')
+        expect(res.text).toContain('>Reported by <')
+        expect(res.text).not.toContain('>Establishment <')
+        expect(res.text).toContain('>Status <')
```
ushkarev, line 315: not _that_ big a deal, this test (and the one below) is quite closely tied to how GDS macros render html. i wonder if we could assert the values we expect in the rows  instead of the column titles? e.g. 
```typescript
…
        expect(res.text).toContain('John Smith')
        expect(res.text).not.toContain('Moorland')
…
```

server/views/pages/dashboard.njk
```diff
@@ -134,7 +140,7 @@
         }) }}
 
         {% if not noFiltersSupplied %}
-          <a href="?" class="govuk-link--no-visited-state" id="clearFilter">Clear</a>
+          <a href="?" class="govuk-link--no-visited-state" id="clearFilter">Clear filters</a>
```
ushkarev, line 143: turns out this needs the `govuk-link` class as well (before the one already there)

server/routes/dashboard/dashboard.ts
```diff
@@ -85,37 +86,12 @@ export default function dashboard(service: Services): Router {
       page,
```
ushkarev, line 86: there is no validation of `location` with respect to user’s caseloads meaning you can see reports you shouldn't

assets/scss/components/_horizontal-content.scss
jonesj2193: That is the value that is used for `@media (min-width:40.0625em) ` throughout the gov css styles. So I just used that

assets/scss/components/_horizontal-content.scss
jonesj2193: Its still setting the overall grey area and setting the background colour the filters are located in. 
And yeah when the errors changed it before so they were on each line and full width, Dan didn't like that and when I showed him it this morning with it staying the same width and in place with the error message, he seemed a lot happier.

ushkarev: in terms of errors, fair enough! but the grey is actually in `.horizontal-form__wrapper` not `.horizontal-form`

assets/scss/components/_horizontal-content.scss
ushkarev: as so you are referring to `@include govuk-media-query($from: desktop)` which is what the scss source is (the css ends up as an actual number)

server/views/pages/dashboard.njk
jonesj2193: When I put in the absolute path it comes up with errors saying it can't find it? tried putting it in from the "server" down and from "views" down. Neither works

server/routes/dashboard/dashboard.ts
```diff
@@ -74,6 +67,15 @@ export default function dashboard(service: Services): Router {
       order = 'DESC'
     }
 
+    if (
+      userRoles.includes(roleReadWrite) &&
+      !userRoles.includes(roleApproveReject) &&
+      !incidentStatuses &&
+      req.url !== '/?'
```
ushkarev, line 74: what is `req.url !== '/?'` checking for?

ushkarev: oh i see, it's when u clear filters. i would imagine that `/` & `/?` should show the same thing and that clearing filters would actually reset them to the default, ie. with to-do checked. don't u think?

ushkarev: in which case you could check for presence or absence of `'incidentStatuses' in req.query` to allow for a deliberately empty selection of checkboxes (which isn't the default)

server/views/pages/dashboard.njk
```diff
@@ -27,10 +27,12 @@
...
+  <h1 class="govuk-heading-l">Incident reports</h1>
+
   {% if permissions.canCreateReport %}
     {% if permissions.canCreateReportInActiveCaseload %}
       {{ govukButton({
-        text: "Report an incident",
+        text: "Create an incident report",
         preventDoubleClick: true,
         href: "/create-report",
         isStartButton: true
```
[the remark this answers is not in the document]
jonesj2193: It wasn't explicitly mentioned. I looked at it when making these changes as it is visually different. I left it as a "start button" as it made sense to me and matched the description on the gov uk website as to what that means/indicates, but it's not something I have explicitly checked.

ushkarev (approved):

ushkarev (approved):

PR #328, opened 2025-02-04
IR-740: Allow jumping to any page in questions form wizard routes

ushkarev:


server/controllers/wip/questionsController.ts
```diff
@@ -64,8 +65,7 @@ export default class QuestionsController extends BaseController<FormWizard.Multi
       const { reportConfig } = res.locals
 
       for (const question of report.questions) {
-        // TODO: Remove QID-stripping logic once removed from API
-        const fieldName = stripQidPrefix(question.code)
+        const fieldName = questionFieldName(question)
```
xoen, line 68: Wouldn't be less opaque to use `question.code` directly? It would also have the benefit that you wouldn't need `questionFieldName()` at all and just use `question.code` or `question.id` as necessary (again without the level of indirection).

ushkarev: both question config and question instances have a `code` property – but they mean very different things so it's safer to not have to remember each time which is which and how in fact you construct a field name

xoen (approved):
Not a fan of `questionFieldName()` but _FINE_, the rest is good tho 👍

PR #334, opened 2025-02-07
IR-740: Allow jumping to any page in questions form wizard routes

ushkarev:


Mjwillis (dismissed):
LGTM

server/data/incidentTypeConfiguration/formWizard.ts
```diff
@@ -47,6 +47,7 @@ export function generateSteps(config: IncidentTypeConfiguration): FormWizard.Ste
         fields,
         controller: QuestionsController,
         template: 'questionPage',
+        entryPoint: true,
```
xoen, line 49: I'm a bit confused by this (and the diff in the test file)...all steps are entry point?

I'm guessing this is the trick that allows jumping to any page of the questionnaire without FormWizard complaining? e.g. all pages could be "entry point", is that right?

ushkarev: yes. i tried a variety of alternatives but this is the one that seemed to require minimal workarounds. i wanna test this out a bit in practice. i'm not 100% convinced it solves all problems, but `QuestionProgress` class still determines a report’s readiness so that should work fine

PR #339, opened 2025-02-12
Fixed RO filter/table bug

jonesj2193:
Updated logic on RO default to not reinstate default when sorting table or moving to another page of table


ushkarev (approved):
that seems to do it

PR #342, opened 2025-02-14
IR-790 Integrate Digital Prison Reporting components and update stati…

Mjwillis:
…c resource handling


server/routes/index.ts
```diff
@@ -63,5 +66,44 @@ export default function routes(services: Services): Router {
   // NOMIS data dumps used for updating constants in this repository
   router.use('/download-report-config', makeDownloadConfigRouter())
 
+  // Digital Prison Reporting
+  get(
```
Mjwillis, line 70: Obviously we wouldn't do it like this! There would be a menu screen etc

server/views/partials/reportsLayout.njk
```diff
@@ -0,0 +1,20 @@
+{% extends "./layout.njk" %}
+
+{% block pageStylesheets %}
+  <link href="/assets/dpr/css/all.css" rel="stylesheet" />
+{% endblock %}
```
ushkarev, line 5: this doesn’t do anything since parent templates don’t specify such a block

package.json
```diff
@@ -84,6 +84,7 @@
   "dependencies": {
     "@aws-sdk/client-sqs": "^3.744.0",
     "@ministryofjustice/frontend": "^3.3.1",
+    "@ministryofjustice/hmpps-digital-prison-reporting-frontend": "^3.37.7",
```
ushkarev, line 87: you don’t need to directly include transitive dependencies since you don’t actually use them directly so the only line you needed to dependencies was this one

server/routes/index.ts
```diff
@@ -63,5 +66,44 @@ export default function routes(services: Services): Router {
...
+    '/incident-reporting/prisoners',
+    ReportListUtils.createReportListRequestHandler({
+      title: 'Prisoners with incidents',
+      definitionName: 'incident-with-people',
+      variantName: 'by-prisoner',
+      apiUrl: config.apis.hmppsIncidentReportingApi.url,
+      apiTimeout: config.apis.hmppsIncidentReportingApi.timeout.deadline,
+      layoutTemplate: 'partials/reportsLayout.njk',
+      tokenProvider: defaultTokenProvider,
```
ushkarev, line 105: is it the token that determines what a user can/cannot see? ie. we are deferring _all_ permissions checks to IRS api and/or DPR?

Mjwillis: Yes at the moment its the API/DPR that decides but I'd prefer this to change in the longer term, at the moment I've made it so that anyone with the read only role can access, but this will also be restricted by caseload etc.

ushkarev:
sorry i evidently lied, you _do_ need the types packages you tried to install: `@types/nunjucks-date` and `@types/parseurl`

it's only to appease type checking which is annoying since it's trying to build a picture of third party code that we don’t touch directly

ushkarev (dismissed):
i think i fixed the stuff i told you to break 🤦‍♂️
…but i can’t test it properly since `dev` gets turned off outside office hours

PR #345, opened 2025-02-17
IR-790: DPR assets

ushkarev:
Continues on from #342 to add missing route to GOV.UK assets for DPR pages and remove seemingly-unnecessary dependency and jQuery route

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,6 @@
     "jwt-decode": "^4.0.0",
     "nocache": "^4.0.0",
     "nunjucks": "^3.2.4",
-    "nunjucks-date": "^1.5.0",
     "passport": "^0.7.0",
     "passport-oauth2": "^1.8.0",
     "redis": "^4.7.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -104,7 +104,6 @@
     "jwt-decode": "^4.0.0",
     "nocache": "^4.0.0",
     "nunjucks": "^3.2.4",
-    "nunjucks-date": "^1.5.0",
```
ushkarev, line 107: it was the types package we need, not the actual package

server/middleware/setUpStaticResources.ts
```diff
@@ -32,9 +32,9 @@ export default function setUpStaticResources(): Router {
     '/node_modules/@ministryofjustice/hmpps-digital-prison-reporting-frontend/dpr/assets',
     '/node_modules/accessible-autocomplete/dist',
   ).forEach(dir => staticRoute('/assets', dir))
-  staticRoute('/assets/js/jquery.min.js', '/node_modules/jquery/dist/jquery.min.js')
```
ushkarev, line 35: we don't use it and DPR doesn't appear to

Mjwillis (approved):
LGTM

PR #346, opened 2025-02-17
Prisoner Summary Page

jonesj2193:
Added prisoner summary page to view existing prisoners linked to a report as per ticket [IR-796]

Can navigate to page from view report screen. If "Continue" is clicked, navigates to "Add a prisoner" journey if option selected is "yes", takes the user back to the "view report" page for that report if option selected is "no" 

[IR-796]: https://dsdmoj.atlassian.net/browse/IR-796?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/controllers/prisonerSummary/prisonerSummary.ts
```diff
@@ -0,0 +1,66 @@
+import type express from 'express'
+import type FormWizard from 'hmpo-form-wizard'
+
+import FormInitialStep from '../base/formInitialStep'
+import type { ReportWithDetails } from '../../data/incidentReportingApi'
+import { prisonerInvolvementOutcomes, prisonerInvolvementRoles } from '../../reportConfiguration/constants'
+
+export default class prisonerSummary extends FormInitialStep {
```
ushkarev, line 8: could you make this extend the `BaseController` class instead pls? it probs should also start with a capital, i'm surprised that eslint didn't get angry!

server/routes/prisonerSummary/fields.ts
```diff
@@ -0,0 +1,5 @@
+import type FormWizard from 'hmpo-form-wizard'
+
+const fields: FormWizard.Fields = {}
```
ushkarev, line 3: shouldn't the yes/no be a field here? otherwise no need for form-wizard if there are no fields…

server/controllers/prisonerSummary/prisonerSummary.ts
ushkarev: it'll mean a bunch of types would need to be more specific, but the upside is better type checking!

jonesj2193: Yeah should be a capital, missed that! will do these now

server/routes/prisonerSummary/fields.ts
jonesj2193: I did try to remove this and all acknowledgement of a fields as the only field is defined in the nunjucks. But it wouldn't pick up the field without naming it in the steps (which is fine), but when I removed this file, then the wizard function in the index complained as it expects to have a "fields" parsed into it and you can't simply remove it. I mean I guess I could remove this file and just parse in a `{}` into the function, which saying that now probably makes more sense

jonesj2193: Also this was the field I had in originally, then had to render via the render fields macro, and the formatting etc was a bit fiddly. that was when you suggested to just put it directly into the nunjucks

server/controllers/prisonerSummary/prisonerSummary.ts
```diff
@@ -0,0 +1,66 @@
...
+        { text: 'Role' },
+        { text: 'Outcome' },
+        { text: 'Details' },
+        { text: 'Action' },
+      ]
+    } else {
+      tableHeading = [{ text: 'Prisoner' }, { text: 'Role' }, { text: 'Details' }, { text: 'Action' }]
+    }
+
+    const backLink = `/reports/${reportId}`
```
ushkarev, line 42: it might be a little better to use form-wizard machinery for the back link. you can add this method instead

```typescript
  getBackLink(_req: FormWizard.Request, res: express.Response): string {
    const reportId = res.locals.report.id
    return `/reports/${reportId}`
  }
```

server/routes/prisonerSummary/index.ts
```diff
@@ -0,0 +1,23 @@
...
+import wizard from 'hmpo-form-wizard'
+
+import { populateReport } from '../../middleware/populateReport'
+import steps from './steps'
+
+const router = express.Router({ mergeParams: true })
+
+router.use(
+  populateReport(true),
+  wizard(
```
ushkarev, line 11: you should add `logoutIf(cannotEditReport)` as well for user permission checks

server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -0,0 +1,110 @@
+{% extends "partials/layout.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "macros/renderFields.njk" import renderFields %}
+
+{% set pageTitle = applicationName + " - Report" %}
```
ushkarev, line 9: we could probably say something like "prisoners involved in report"…?

server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -0,0 +1,110 @@
...
+  {% endfor %}
+
+  {{ govukTable({
+    head: tableHeading,
+    rows: rows
+  }) }}
+
+{% endif %}
+
+<form id="form-wizard" method="post" novalidate>
```
ushkarev, line 58: doesn't hugely matter since there are no references to this form id, but it's a little generic? i'd either not give it an id at all or make it more related prisoner-involvements

server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -0,0 +1,110 @@
...
+          { text: prisonerInvolvementLookup[prisoner.prisonerRole] or prisoner.prisonerRole },
+          { text: prisoner.comment },
+          { html: "<a href='/reports/" + report.id + "'>Remove</a>" + ' | ' + "<a href='/reports/'>Edit</a>" }
+        ]
+        ), rows) %}
+    {% endif %}
+  {% endfor %}
+
+  {{ govukTable({
+    head: tableHeading,
```
ushkarev, line 52: i would add a class, one that will be shared with the staff involvement table, like `app-involvement-table` or similar. because looking at the designs, we will need to add a style to right-alight the last table header cell:
```scss
.app-involvement-table {
  th:last-child {
    text-align: right;
  }
}
```

server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -0,0 +1,110 @@
+{% extends "partials/layout.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "macros/renderFields.njk" import renderFields %}
+
+{% set pageTitle = applicationName + " - Report" %}
+{% set bodyClasses = "app-view-report" %}
```
ushkarev, line 10: we don’t need the particular changes this class does for this page, but i'm thinking… maybe it _is_ a nice idea to give all pages unique class names? that way, we can make unit/integration tests know for sure what template was rendered 🤔
or just remove this line for now 😆

server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -0,0 +1,110 @@
+{% extends "partials/layout.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "macros/renderFields.njk" import renderFields %}
```
ushkarev, line 7: don’t need the `renderFields` import anymore

server/routes/index.ts
```diff
@@ -37,6 +38,7 @@ export default function routes(services: Services): Router {
   router.use('/reports/:id', viewReportRouter(services))
   router.use('/reports/:id/update-details', updateDetailsRouter)
   router.use('/reports/:id/questions', questionsRouter)
+  router.use('/reports/:id/prisoners', prisonerSummaryRouter)
```
ushkarev, line 41: soz i've changed the param name in #347 so you’ll need `:reportId` here now 🙂

ushkarev (approved):

PR #351, opened 2025-02-19
Added staff summary page. Also minor tweak to prisoner summary page.

jonesj2193:


server/views/pages/prisonerSummary/prisonerSummary.njk
```diff
@@ -55,6 +55,8 @@
       rows: rows
     }) }}
 
+  {% else %}
+    <p>No prisoners have been added to this incident report.</p>
```
ushkarev, line 59: not that i'm against this, but i can't see it in designs…?

ushkarev (approved):

server/views/pages/prisonerSummary/prisonerSummary.njk
jonesj2193: As mentioned in slack, it is "in" the designs, just not explicitly shown on a page.

PR #353, opened 2025-02-21
IR-798: Add & edit prisoner and staff involvements

ushkarev:
Adds pages where prisoner & staff involvements can be added or changed – roles, comments etc.


server/routes/reports/prisoners/involvement/controller.ts
```diff
@@ -0,0 +1,75 @@
...
+  locals(req: FormWizard.Request<Values>, res: express.Response): Partial<FormWizard.Locals<Values>> {
+    return {
+      ...super.locals(req, res),
+      pageTitle: `${possessive(nameOfPerson(this.getPrisonerName(res)))} involvement in the incident`,
+    }
+  }
+
+  protected errorMessage(error: FormWizard.Error): string {
+    if (error.key === 'prisonerRole') {
+      return 'Chose the prisoner’s role'
```
xoen: Shouldn't this be 'choose' (two 'o's)? (same below for the outcome field)

xoen: Ops, I now realise it was fixed in a later commit, soz

ushkarev: fixed in a later commit!

xoen (approved):

PR #355, opened 2025-02-24
IR-790 Refactor date formatting functions to use longDate

Mjwillis:


server/utils/nunjucksSetup.ts
```diff
@@ -94,7 +94,7 @@ export default function nunjucksSetup(app: express.Express): void {
 
   // date/datetime formatting
   njkEnv.addFilter('dateAndTime', format.dateAndTime)
-  njkEnv.addFilter('date', format.date)
+  njkEnv.addFilter('longDate', format.longDate)
```
xoen, line 97: I think the tests fail because of one the `date` filter still used [in the `reports/view.njk` template](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/beffdce8c218a618f92d7ea0f4e33b10c921501d/server/views/pages/reports/view.njk#L32)

Mjwillis: Corrected now

xoen (approved):

PR #356, opened 2025-02-25
IR-797: Involvement search

ushkarev:
The full journey to add, edit and remove prisoner and staff involvements should be working.

Caveat: back links, continue and exit buttons will generally make more sense in the context of editing a report – the create journey should all tie together but that will be part of a different PR.


jonesj2193 (approved):
Makes sense to me overall. Will look to play about with it more and understand flows etc, where they are how they works, and test it fully once merged in.

xoen (approved):

PR #357, opened 2025-02-25
IR-799 remove prisoner

jonesj2193:
Added remove prisoner user journey. 

Also added partial template to use the govuk notification banner - currently only set to use the success banner option.


server/data/incidentReportingApi.ts
```diff
@@ -173,6 +173,7 @@ export type StaffInvolvement = {
 }
 
 export type PrisonerInvolvement = {
+  sequence?: number
```
ushkarev, line 176: this field is deprecated, we shouldn't really refer to it

server/controllers/removePrisoner/removePrisoner.ts
```diff
@@ -0,0 +1,69 @@
...
+    const locals = super.locals(req, res)
+    const { errors } = res.locals
+    const report = res.locals.report as ReportWithDetails
+    const { index } = req.params
+
+    if (errors.removePrisoner) {
+      errors.removePrisoner.message = 'Select if you would like to remove this prisoner to continue.'
+    }
+
+    const prisonerToRemove = report.prisonersInvolved[Number(index) - 1]
```
ushkarev, line 27: it’d be better to use `parseInt(index, 10)` here and in other places below: casting to a number is a bit too lenient

server/routes/reports/prisoners/remove/removePrisoner.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    return request(app)
+      .get(viewReportUrl)
+      .expect('Content-Type', /html/)
+      .expect(res => {
+        expect(res.text).toContain('Are you sure you want to remove A1111AA: Andrew Arnold?')
+        expect(res.text).toContain('Yes')
+        expect(res.text).toContain('No')
+        expect(res.text).toContain('Continue')
+        expect(res.text).toContain('Save and exit')
+        expect(incidentReportingApi.getReportWithDetailsById).toHaveBeenCalledTimes(1)
```
ushkarev, line 58: we should assert that `deleteFromReport` is called with the correct value.

you'd need a little bit of mock setup, though. check out the use of `incidentReportingRelatedObjects ` here:
https://github.com/ministryofjustice/hmpps-incident-reporting/pull/353/commits/dddc665c90d554226066199e07a563229277fdd1#diff-4b826d58631e95f8b773a3121046c0882e3b68fc65609f26020d1401c61cf37bR35-R37

server/data/incidentReportingApi.ts
jonesj2193: I realised I had left those in as I committed it, the code doesn't use it anymore, so I will remove from types etc.

ushkarev: in fact i don’t think you ever read it?

ushkarev (approved):

PR #358, opened 2025-02-25
IR-820: Prisoner involvement roles

ushkarev:


Mjwillis (approved):
LGTM

jonesj2193 (approved):
LGTM. Wasn't necessarily aware of the roles being unique in some report types. But if it's a requirement. Makes sense

ushkarev:
@jonesj2193 yeah it's something that nomis does and the ticket has:

> Role values are dynamic based on incident type, see attached Incidents_Role-Involvement_LOVs - respective LOVs will be presented to the user based on the incident type selected

PR #362, opened 2025-02-26
IR-820: Show error when there are no prisoner roles available

ushkarev:
Screenshot 2025-02-27 at 12 33 41
accidentally missed in #358 – there is a situation where you try to add or edit a prisoner involvement but there are no roles available


jonesj2193 (approved):
LGTM

PR #364, opened 2025-03-03
IR-820: Allow extra prisoner roles as options for 3 incident types

ushkarev:
app-involvement-request


xoen (approved):
Aren't these going to be overwritten when we convert the reports again?

I suppose the comments are enough of an hint they shouldn't be removed.

ushkarev:
yes indeed they could be overwritten accidentally: that's why i wanted to put in the comments so it shows more obviously in a diff

xoen:
A test is an option, but I don't mind as it is, we are usually checking these reports configurations when committing them and on top of that there is a review process so that's probably enough 👍

PR #370, opened 2025-03-09
IR-954: Allow changing a report’s type

ushkarev:
Depends on [api#196](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/196): API handles archiving questions and clearing prisoner involvements.

### DIFF
--- a/server/views/pages/reports/confirm-type-change.njk
+++ b/server/views/pages/reports/confirm-type-change.njk
@@ -1,0 +1,12 @@
+{% extends "partials/formWizardLayout.njk" %}
+
+{% set pageTitle = "Some of your answers will be deleted" %}
+{% set bodyClasses = "app-confirm-change-type" %}
+
+{% block formFields %}
+  <p>
+    Changing the type of incident will remove some of your answers.
+    {# TODO: it's not the summary that needs re-entering #}
+    You’ll need to re-enter the incident summary and any prisoners that were involved.
+  </p>
+{% endblock %}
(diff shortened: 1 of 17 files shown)


Mjwillis (dismissed):
LGTM

server/views/pages/reports/confirm-type-change.njk
```diff
@@ -0,0 +1,12 @@
+{% extends "partials/formWizardLayout.njk" %}
+
+{% set pageTitle = "Some of your answers will be deleted" %}
+{% set bodyClasses = "app-confirm-change-type" %}
+
+{% block formFields %}
+  <p>
+    Changing the type of incident will remove some of your answers.
+    {# TODO: it's not the summary that needs re-entering #}
+    You’ll need to re-enter the incident summary and any prisoners that were involved.
```
jonesj2193, line 10: I didn't think the summary did need re-entering? I can see the TODO message so assume this is acknowledged to be wrong. Me and Steve have commented in the figma on the content needing to be reviewed and updated for this page to represent what will change/need to be done.

ushkarev: yup. designs still have this text; that's why i left the TODO msg

xoen: (on a similar note, **all** answers will be removed, not just "some of your answers" as the questions set is different when changing the time)

jonesj2193: This is what me and Steve have said in the figma. So I think they are reviewing what it will be

PR #374, opened 2025-03-11
IR-824 DPR reporting enhancements

Mjwillis:
- Add callback to setup routes using system token 
 - Add correct breadcrumbs 
 - upgrade DPR UI library to fix date issue
 - Add basic integration tests


server/routes/dpr/index.ts
```diff
@@ -24,12 +25,18 @@ export function dprRouter(): Router {
...
+            },
           }),
         )
       }
     }
+  })
 
+  router.get('/', async (req, res) => {
+    const { incidentReportingApi } = res.locals.apis
+    const definitions = await incidentReportingApi.getManagementReportDefinitions()
```
Mjwillis, line 39: We get the definitions again because we want DPR to apply the policies for THIS logged in user to know if the report is accessible

server/routes/dpr/index.ts
```diff
@@ -24,12 +25,18 @@ export function dprRouter(): Router {
             apiUrl: config.apis.hmppsIncidentReportingApi.url,
             apiTimeout: config.apis.hmppsIncidentReportingApi.timeout.deadline,
             layoutTemplate: 'partials/dprLayout.njk',
-            tokenProvider: defaultTokenProvider,
+            tokenProvider: (req: Request, res: Response, next: NextFunction) => {
+              return res.locals.systemToken
```
Mjwillis, line 29: overridden the default token provider as we don't want the user token but the system token with the username added.

There is more work to be done on the API side to use this token correctly to look up the users roles - this is future work but for now it works because I've add the roles to the system token (which actually are user roles) - these will be removed in future once API is corrected

server/views/pages/managementReporting/index.njk
```diff
@@ -30,15 +30,17 @@
   <p>Below is the of available management reports</p>
 
   {% for definition in definitions %}
+    {% if definition.authorised %}
     <h2 class="govuk-heading-m">{{ definition.name }}</h2>
     <p class="govuk-body">{{ definition.description }}</p>
 
     {% for variant in definition.variants %}
-      <h3 class="govuk-heading-s"><a href="/management-reporting/{{ definition.id }}/{{ variant.id }}" class="govuk-link" target="_blank"> {{ variant.name }} </a></h3>
+      <h3 class="govuk-heading-s"><a href="/management-reporting/{{ definition.id }}/{{ variant.id }}" class="govuk-link"> {{ variant.name }} </a></h3>
```
Mjwillis, line 38: Now we have the breadcrumbs we can remove the target = _blank

server/views/pages/managementReporting/index.njk
```diff
@@ -30,15 +30,17 @@
   <p>Below is the of available management reports</p>
 
   {% for definition in definitions %}
+    {% if definition.authorised %}
```
Mjwillis, line 33: Check if the user is authorised to view report - at the moment this will always return true

ushkarev (approved):

ushkarev (approved):

PR #380, opened 2025-03-12
IR-68: Update app for MoJ frontend v4

ushkarev:
…where they started exporting the filters as part of the package [moj-frontend#1256](https://github.com/ministryofjustice/moj-frontend/pull/1256)

Related to #377


Mjwillis (approved):
LGTM

PR #386, opened 2025-03-17
IR-740: Fix question/response configuration traversal

ushkarev:
Key lock incident highlighted a few problems related to inactive question & response configuration. This was not always filtered out meaning that
- some valid responses were accidentally being deleted
- the incorrect next question/page was being chosen

Also, clearing form wizard session when questions are saved successfully in the hope of preventing form wizards interfering with each other.

### DIFF
--- a/server/services/questionsToDelete.ts
+++ b/server/services/questionsToDelete.ts
@@ -96,7 +96,7 @@ export default class QuestionsToDelete {
     answeredQuestions: AnsweredQuestion[],
   ): string | symbol | null {
     const answeredQuestion = answeredQuestions.find(question => question.code === start)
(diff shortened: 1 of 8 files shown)


server/services/questionsToDelete.ts
```diff
@@ -106,15 +106,16 @@ export default class QuestionsToDelete {
         answerConfig = findAnswerConfigByCode(response.response, questionConfig)
       }
     } else {
-      // Current question not answered, but peraphs there is no branching...
+      // Current question not answered, but perhaps there is no branching...
```
xoen, line 109: I love this word but I always struggle to spell it 😅

xoen (approved):
Nice work 👍

PR #388, opened 2025-03-17
IR-943 Temporarily allow Dev access in prisons for Rose's prison visit

Mjwillis:


ushkarev (commented):
`allowlist: null` means there are no IP restrictions present in this app. however, HMPPS auth won't let you log in from elsewhere even in `dev`

PR #390, opened 2025-03-21
IR-948: Improve performance of questions grouping

xoen:
- keep track of which steps were grouped to the limit and can't be grouped further
- when the `groupStepsStartingAt()` is asked to group these steps again it knows it's already grouped so it just returns quickly without attempting to group it again

Because of the nature of the reports config and
the various branching some parts of the questionnaire may be reached via different routes which means the grouping logic may attempt to traverse the same section of the graph multiple times.

This is an optimisation that prevents doing the same work more than once when we know that subtree can't be grouped any further (think memoisation but with the key difference that it is perfectly fine to invoke `groupStepsStartingAt()` with the same step because of the side effect of grouping)

NOTE: This should **not** change the shape of the questionnaires or how the questions are grouped. End result of the grouping should be exactly the same in theory.


server/data/testData/ESCAPE_FROM_CUSTODY.ts
```diff
@@ -0,0 +1,6214 @@
...
+    },
+    {
+      prisonerRole: 'SUSPECTED_INVOLVED',
+      onlyOneAllowed: false,
+      active: true,
+    },
+  ],
+} as const
+
+export const steps: FormWizard.Steps<FormWizard.MultiValues> = {
```
xoen, line 4835: FYI: I've also checked this whole expected `steps` dictionary manually yesterday against the SVG graph and how the grouping rules work [as you can imagine that is a lot of "fun" 😅] but feel free to triple check or spot check this if you wish

ushkarev (approved):

PR #394, opened 2025-03-25
IR-940: Warn users if incident date is over a year in the past during create journey

ushkarev:


Mjwillis (approved):
LGTM

server/utils/parseDateTime.ts
```diff
@@ -0,0 +1,35 @@
+// NB: this module is shared with frontend code!
```
xoen, line 1: perhaps this module could be names `utils/dateUtils`?

ushkarev: There is date formatting in the `format` module so maybe that'd be too generic

xoen (approved):

server/utils/parseDateTime.ts
xoen: Interesting the `format` module is pretty much all to do with formatting dates and times 😅 

...while this one is all about parsing dates and times from user input, so I see why you named it `parseDateTime` 👍 (to me `parseDateTime` sounds almost like an utility function to parse datetimes IYKWIM rather than a noun thing but it's fine, having this in its own module makes sense and we can always rename in the future if we really want to)

PR #402, opened 2025-03-31
IR-68: Fixed esbuild setup from upstream template

ushkarev:


esbuild/esbuild.config.js
```diff
@@ -9,6 +9,21 @@ const buildApp = require('./app.config')
...
 
+/**
+ * Simple debounce helper
+ * @param {Function} fn - The function to debounce
+ * @param {number} delay - The delay in ms
+ * @returns {Function}
+ */
+function debounce(fn, delay = 200) {
+  /** @type {number} */
+  let timeout
```
xoen, line 20: So this is sharing the timeout variable and if `debounce()` is called multiple times in rapid succession the `fn` function is invoked once the last time after 200ms?

xoen: isn't the `timeout` inside this function shared across the various invocations because of the the closure? Presumably that's intended. Or I'm reading this wrong.

ushkarev: yup. it is a GOD SEND copied over from the template project. right now, the local dev server takes like half a minute to start on my machine, but is infinitely faster with this change

xoen: yeah I'll take that, as I've mentioned before sometimes I've seen really weird resource usage, especially when changing Git branch while the dev server is still running. And the idea makes sense 👍

PR #403, opened 2025-03-31
IR-964: Question page validation

ushkarev:
…this gives question pages bespoke error messages quoting the specific question – so that if there are multiple errors on a page, they can be distinguished


jonesj2193 (approved):

server/data/incidentTypeConfiguration/utils.ts
```diff
@@ -16,6 +16,29 @@ export function conditionalFieldName(
   return `${questionFieldName(question)}-${answer.id}-${suffix}`
 }
 
+export function parseFieldName(
```
xoen, line 19: So this is now assuming that questions/answers' IDs are numbers?

At the moment in the questionnaire configuration these are strings which give us potentially more flexibility. But we also use the dash in the forms to name inputs.

Unless I'm confusing the two.

Maybe at some point we formalise this in the configurations to be numbers too? Not to be part of this, but just thinking about potential issues if we add some (now) "weird" id for questions/answers down the line (e.g. a dash in the `id` itself).

ushkarev: that's a v good point. shall i leave a to-do message in a couple of places? e.g. id must not have `-` and in the parser to say that the ids don't actually have to be numbers cos they're _used as strings_!

xoen: I guess it depends on how we want to move forward with these IDs. I think it could be _**potentially**_ nice to be able to name questions better than just `12345` in future questionnaires...in that case maybe this utility shouldn't try and parse numbers but just split by `-`?

But if we want to be more strict and say ids are numbers then probably we'd have to change the incident types configurations type and configs accordingly...(probably not my favourite option tbh)

If we stick to strings for IDs perhaps at some point we can add a validation to check that IDs do not contain `-`? (as I'd imagine this could cause problems with newly introduced utility function).

xoen (commented):
Not sure if we should parse the first 2 parts of a field name as numbers given these could feasibly be strings containing non-digits (and that's ok).

Maybe just split by `-` and we could at some point add a validation against questions/answers with a `-` (which can cause issues with the splitting)

server/data/incidentTypeConfiguration/utils.ts
xoen: Regarding the TODO message: I don't think this is something we just swipe under a TODO comment.

Limiting the IDs to numbers is arbitrary and new to this utility function. I would personally just split by `-` and return the string parts as-is.

Then we can add a validation on the config IDs to highlight when any of these have `-` which can break this.

xoen (approved):

PR #406, opened 2025-04-01
IR-959: Type families

ushkarev:
Depends on [api#299](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/299)

- incident types & families imported from api
- type codes have all been renamed
- type descriptions come from their families
- filtering in the dashboard does NOT yet list or filter by families
- type codes are NOT yet shown to users anywhere


Mjwillis (dismissed):
LGTM

server/reportConfiguration/constants/typeHints.ts
```diff
@@ -5,9 +5,9 @@ import type { Type } from './types'
 /** Additional info to display  */
 // eslint-disable-next-line import/prefer-default-export
 export const typeHints: Partial<Record<Type, string>> = {
```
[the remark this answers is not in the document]
ushkarev: yeah that's right. i actually improve the JSDoc comment in a later commit pointing out that this is for _types_ not _families_ (for now)

server/reportConfiguration/types/ATTEMPTED_ESCAPE_FROM_PRISON_1.ts
```diff
@@ -5439,7 +5439,7 @@ export const ATTEMPTED_ESCAPE_FROM_PRISON_1: IncidentTypeConfiguration = {
       active: true,
       code: 'HOW DID THE PRISONER ATTEMPT TO GAIN ACCESS TO THE EXTERNAL PERIMETER',
       label: 'How did the prisoner attempt to gain access to the external perimeter?',
-      multipleAnswers: true,
```
xoen, line 5442: 👍

server/reportConfiguration/types/ATTEMPTED_ESCAPE_FROM_PRISON_1.ts
```diff
@@ -4431,7 +4431,7 @@ export const ATTEMPTED_ESCAPE_FROM_PRISON_1: IncidentTypeConfiguration = {
       active: true,
       code: 'HOW DID THE PRISONER ATTEMPT TO BREACH THE EXTERNAL PERIMETER',
       label: 'How did the prisoner attempt to breach the external perimeter?',
-      multipleAnswers: true,
```
xoen, line 4434: 👍

server/reportConfiguration/types/FIND_6.ts
```diff
@@ -3261,7 +3261,7 @@ export const FIND_6: IncidentTypeConfiguration = {
           label: 'Roach',
           commentRequired: false,
           dateRequired: false,
-          nextQuestionId: '67214',
+          nextQuestionId: '67213',
```
xoen, line 3264: 👍

server/reportConfiguration/types/RELEASE_IN_ERROR_1.ts
```diff
@@ -1020,7 +1020,7 @@ export const RELEASE_IN_ERROR_1: IncidentTypeConfiguration = {
       active: true,
       code: 'Has the person been returned to custody?',
       label: 'Has the person been returned to custody?',
-      multipleAnswers: true,
+      multipleAnswers: false,
```
xoen, line 1023: 👍

server/data/incidentTypeConfiguration/conversion.ts
```diff
@@ -104,41 +106,30 @@ export function toTypescript({
...
+digraph {{ config.incidentType }} {
+  rankdir=LR;
+  node [shape = circle];
+
+  START_NODE [label="", shape=none];
+  START_NODE -> {{ config.startingQuestionId }} [label = "start"];
+  END_NODE [label="END", shape="doublecircle"];
+
+  {%- for question in questions %}
+  {{ question.id }} [label=< <FONT COLOR="royalblue">{{ question.id }} </FONT> {{ question.label }} >{% if not question.active %}, style="filled", color="#DDD"{% endif %}];
```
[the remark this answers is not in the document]
ushkarev: yeah that's a good idea. i will check if graphviz cares about indentation. i won't regenerate .dot or .svg files cos they won't change in any meaningful way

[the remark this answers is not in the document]
xoen: > i will check if graphviz cares about indentation
Good point, _from memory_, I don't think graphviz cares about the indentation but I may be wrong.

server/data/incidentTypeConfiguration/conversion.ts
```diff
@@ -104,41 +106,30 @@ export function toTypescript({
...
-    for (const answer of question.answers) {
-      const answerStyle = answer.active === true ? '' : inactiveStyle
-
-      const nextNode = answer.nextQuestionId ?? 'END_NODE'
-
-      result += `  ${question.id} -> ${nextNode} [label = < <FONT COLOR="royalblue">${answer.id} </FONT> ${replaceAngleBrackets(answer.label)} >${answerStyle}];\n`
-    }
-  }
-
-  result += '}\n'
```
xoen, line 135: Is this curly brace `}` gone 🤔?

server/data/incidentTypeConfiguration/conversion.ts
```diff
@@ -104,41 +106,30 @@ export function toTypescript({
...
+  START_NODE -> {{ config.startingQuestionId }} [label = "start"];
+  END_NODE [label="END", shape="doublecircle"];
+
+  {%- for question in questions %}
+  {{ question.id }} [label=< <FONT COLOR="royalblue">{{ question.id }} </FONT> {{ question.label }} >{% if not question.active %}, style="filled", color="#DDD"{% endif %}];
+  {%- for answer in question.answers %}
+  {{ question.id }} -> {{ answer.nextQuestionId or 'END_NODE' }} [label=< <FONT COLOR="royalblue">{{ answer.id }} </FONT> {{ answer.label }} >{% if not answer.active %}, color="#DDD"{% endif%}];
+  {%- endfor %}
+  {%- endfor %}
 }
```
xoen, line 128: ah, the closing curly brace of the TS `for` is not in the string 👍 (I don't find nunjucks particularly readable tbh)

ushkarev: no i agree, it's not exactly readable. but i tried to escape html chars myself and was hitting issues somewhere in graphviz.

i almost went for a worse solution: replacing angle brackets with similar-looking small angle brackets
`<` to `﹤` but it looked really weird in svg

server/reportConfiguration/constants/typeHints.ts
```diff
@@ -14,3 +16,49 @@ export const typeHints: Partial<Record<Type, string>> = {
     'Includes suspected and reported self-harm. Do not use to report a noose, unless it’s around the neck or applying pressure.',
   TOOL_LOSS_1: 'Do not use for radio and key or lock compromises. They are separate incident types.',
 }
+
+/**
+ * Short titles displayed when generating titles in the form “About the incident”, eg. “About the assault”.
+ * A *family* of types shares one title.
+ */
+export const shortTypeTitles: Partial<Record<TypeFamily, string>> = {
```
xoen, line 24: Isn't this the lower-cased family description on the API side? Any reason to hardcode those? Maybe there is some special exception here I'm missing?

ushkarev: they weren't the same initially in the content spreadsheet i took this from. i decided to still keep this map in case the words diverge again. it's a bit more explicit even if it's duplication of text 🤷

xoen: Why would they need to differ tho? And on the API side (in the title generation) the rule is simpler so I'd argue it would be consistent to do the same here.

But that's just my 2 cents.

ushkarev: i agree, they don't _need_ to be different. but even now, not all match! check this one out:
family code: `MOBILE_PHONE`
type & family description: “Mobile phone”
question page title: “About the mobile phone incident” (surprise!)

xoen: OK, so **there are** differences. In this case fair enough, they need to be ad-hoc.

(potentially we could have a slightly clever fallback mechanism where the string can be overridden when needs to be different, or default/fallback to the lower case version. That would have the added benefit of highlighting these differences when they're there...but I think that may be overkill and this is totally fine 👍)

xoen (approved):

PR #414, opened 2025-04-08
IR-979 & IR-947: Validate and submit a draft report for review

ushkarev:
if the user can edit a report and it is a draft, they can now submit it for review.

other statuses that *will* become submittable are not covered until the big status rename happens. lifecycle needs confirmation first.


server/routes/reports/viewReport.test.ts
```diff
@@ -461,77 +746,110 @@ describe('View report page', () => {
...
-          const warningShows = res.text.includes('This report can only be amended in NOMIS')
-          expect(warningShows).toBe(warn)
-        })
+      it(`should ${canEdit ? 'warn' : 'not warn'} that report is only editable in NOMIS`, () => {
+        config.activePrisons = ['LEI']
+
+        return request(appWithAllRoutes({ services: { userService }, userSupplier: () => user }))
+          .get(viewReportUrl)
+          .expect(res => {
+            const warningShows = res.text.includes('This report can only be amended in NOMIS')
```
jonesj2193, line 824: Which reports can only be amended in NOMIS? I wasn't aware this was a thing and thought users could edit all reports in the new service?

jonesj2193 (approved):
all looks good and makes sense. Comment added with question more for my knowledge rather than blocking merge.

server/routes/reports/viewReport.test.ts
ushkarev: it's the reports whose locations are not yet live in this service. since we do not sync back to nomis, we naturally cannot allow a report to be editable in both at the same time.

(i didn't add that as part of this PR; https://github.com/ministryofjustice/hmpps-incident-reporting/pull/304/commits/ed0809142e54302992147049ee59da9bc513d44d)

xoen (approved):

PR #418, opened 2025-04-11
IR-973: Clear form wizard session more strategically

ushkarev:
Form wizard sessions were being cleared slightly too early. The root controller class’ success handler [adds a history step to the journey model](https://github.com/HMPO/hmpo-form-wizard/blob/f9ad4df65d500eea2128e6773f09980e17e53d3b/lib/controller/index.js#L360) _after_ we cleared the session. It’s presence in the session later meant those same controller could not tell where to send the user.

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -66,6 +66,8 @@ export declare global {
       reportSubUrlPrefix?: string
       /** Some routes load incident type config */
       reportConfig?: IncidentTypeConfiguration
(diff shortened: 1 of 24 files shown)


Mjwillis (approved):

Mjwillis:
Looks ok - probably needs a bit of heavy testing once deployed

ushkarev:
yes, totally does! i mean i didn't even spot the problem which means i never tried to create 2 reports in a row

PR #425, opened 2025-05-01
Ir 1003 view addendums

jonesj2193:
added viewing description addendums to view report page


server/data/testData/incidentReporting.ts
```diff
@@ -115,6 +118,24 @@ export function mockReport({
   if (withDetails) {
     return {
       ...basicReport,
+      descriptionAddendums: withAddendums
+        ? [
+            {
+              createdBy: 'user1',
+              createdAt: format.isoDateTime(new Date('2023-12-05T15:34:00')),
```
ushkarev, line 125: i think it's a better idea to base these dates on `incidentDateAndTime` so that the addenda aren't in the past

server/views/pages/reports/view/_details.njk
```diff
@@ -1,5 +1,30 @@
...
+    <br />
+    <small>{{ report.reportedAt | timeOnDate }}</small>
+    <br />
+    {{ report.description | escape | nl2br }}
+    {% for addendum in report.descriptionAddendums %}
+      <br />
+      <br />
+      <strong>{{ addendum | nameOfPerson }}</strong>
+      <br />
+      <small>{{ addendum.createdAt | timeOnDate }}</small>
```
ushkarev, line 16: `timeOnDate` isn't a real filter

server/utils/format.ts
```diff
@@ -20,6 +20,28 @@ export default {
     return formatted.replace(' at ', ', ')
   },
 
+  /**
+   * Format time of day on Date as Europe/London.
+   *
+   * Example: `11:00 on 22 February 2022`
+   */
+  timeOnDate(date: Date): string {
```
ushkarev, line 28: why do we need a new date format? i don't think that we should show dates in a different way on every page

server/data/incidentReportingApi.ts
```diff
@@ -136,6 +137,14 @@ export type PaginationSortingParams = {
   sort: string[]
 }
 
+export type DescriptionAddendum = {
+  createdBy: string | null
```
xoen, line 141: This is not nullable, the field is always populated (but for some records it may be the SYSTEM USERNAME)

server/views/pages/reports/view/_details.njk
```diff
@@ -1,5 +1,30 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 
+{% if report.descriptionAddendums.length != 0 %}
+
+  {% set descriptionHtml %}
+    <strong>{{ usersLookup[report.reportedBy].name or report.reportedBy }}</strong>
+    <br />
+    <small>{{ report.reportedAt | timeOnDate }}</small>
```
ushkarev, line 8: i would recommend adding a class to make the font size stable `govuk-body-s` feels appropriate, but do check the padding and margins! otherwise maybe `govuk-!-font-size-16`

https://design-system.service.gov.uk/styles/paragraphs/#body-small
https://design-system.service.gov.uk/styles/font-override-classes/#font-size

server/routes/reports/viewReport.test.ts
```diff
@@ -863,3 +864,47 @@ describe('View report page', () => {
     })
   })
 })
+
+describe('View report page with description addendums', () => {
```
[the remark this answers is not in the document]
jonesj2193: Yeah makes sense, I pasted it from another test then kinda forgot to go back and trim it down etc.

server/utils/format.ts
xoen: I am with Igor on this one, I would guess the design is probably shown a different format "by accident"?

ushkarev (approved):
well pending the design answer regarding random date formatting, i suppose we can merge this anyway

PR #431, opened 2025-05-09
IR-1027: Backport template changes

xoen:
Main changes:
  - Use classes from HMPPS libraries `@ministryofjustice/hmpps-rest-client` and `@ministryofjustice/hmpps-auth-clients`
  - Use `RestClient` from `@ministryofjustice/hmpps-rest-client`
  - Use `RedisTokenSTore`/`InMemoryTokenStore` from `@ministryofjustice/hmpps-rest-client`
  - Use `VerificationClient` from `@ministryofjustice/hmpps-rest-client` (instead of custom `tokenVerifier()` function)
  - Use `AuthenticationClient` from `@ministryofjustice/hmpps-rest-client` (instead of custom `HmppsAuthClient`)


server/data/incidentReportingApi.test.ts
```diff
@@ -1,7 +1,7 @@
 import nock from 'nock'
+import { SanitisedError } from '@ministryofjustice/hmpps-rest-client'
 
 import config from '../config'
-import type { SanitisedError } from '../sanitisedError'
```
xoen, line 4: Bear with me as I thought I also got rid of the `SanitisedError` interface but I didn't which is a bit confusing as it's slightly different from the type in the library (library one has `responseStatus` not `status`).

server/data/testData/thrownErrors.ts
```diff
@@ -8,8 +8,8 @@ import { SanitisedError } from '../../sanitisedError'
  */
 // eslint-disable-next-line import/prefer-default-export
 export function mockThrownError<T>(responseBody: T, status: number = 400): SanitisedError<T> {
-  const error = new SanitisedError<T>(`Error: ${status}`)
```
ushkarev, line 11: `SanitisedError` is a real class extending error so sure we'd want to instantiate the real thing still?

server/middleware/populateReport.test.ts
```diff
@@ -90,8 +91,6 @@ describe('report-loading middleware', () => {
     expect(res.locals.report).toBeUndefined()
     expect(res.locals.reportUrl).toBeUndefined()
     expect(res.locals.reportSubUrlPrefix).toBeUndefined()
-    expect(next).toHaveBeenCalledWith(
-      expect.objectContaining({ message: 'populateReport() requires req.params.reportId', status: 501 }),
-    )
+    expect(next).toHaveBeenCalledWith(new NotImplemented('populateReport() requires req.params.reportId'))
```
ushkarev, line 94: how come this one needs to be an instance of an error and the others are still just checking for message text and status number?

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "strict": false,
     "module": "CommonJS",
     "moduleResolution": "Node",
+    "isolatedModules": true,
```
ushkarev, line 9: fyi: this makes unit tests noticeably slower :(

server/data/testData/thrownErrors.ts
xoen: That's what I tried at first, but I'm getting a bunch of `TypeError: hmpps_rest_client_1.SanitisedError is not a constructor`

ushkarev: i wonder if those errors happened before you swapped in the `SanitisedError` from the lib? cos the one in the app before was not a constructor… but surely this one is cos it's just a class… :/

server/middleware/populateReport.test.ts
xoen: oh, I think it's because `populateReport()` throws a `NotImplemented` error directly which is not a `SanitisedError`

xoen: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/431/commits/08ac8b828022a0eb6140b024117264f7aaa3856e

server/data/testData/thrownErrors.ts
xoen: Maybe it's because [it's exported as a `Type`](https://github.com/ministryofjustice/hmpps-typescript-lib/blob/0ae5651e0d8e0b44e6eeacfc3b108fc9fb1375be/packages/rest-client/src/main/index.ts#L10) rather than as a class?

ushkarev: that's likely the problem. also, i take it back cos there's no real class instantiated here anyway:
https://github.com/ministryofjustice/hmpps-typescript-lib/blob/6fa730c39882f4b6411fc81f2bea38b2be69a6e8/packages/rest-client/src/main/helpers/sanitiseError.ts#L8

ushkarev (approved):

PR #437, opened 2025-05-15
IR-1009: Fix glaring spelling mistakes

ushkarev:
…without worrying about clarity of meaning or grammar at this stage.

Machine codes are unaffected; this is only what people see on this site.


server/utils/utils.ts
```diff
@@ -237,7 +238,7 @@ export function convertToSentenceCase(str: string): string {
     'NOMS',
     'NOU',
     'NPS',
-    'NTRG',
+    'NTRG', // TODO: is it NTRG or NRTG?
```
xoen, line 241: I think it's referring to NTRG (National Tactical Response Group) but I may be wrong. When I bumped into this I looked it up [in acronyms](https://ministryofjustice.github.io/acronyms/) and it made sense in sentences like:

> NTRG And / Or NDTSG Assistance Requested

xoen (approved):

server/utils/utils.ts
ushkarev: aah @xoen so should i change the “Nrtg” i found do u think?

xoen: It's a tricky one as I am not 100% sure of course. I see why we would want to have both the same as they seem to be the same answer. But I don't know if this is changing the semantic of the answer?

I think they probably mean the same thing so it's ok to change, we can always change back if this was deliberate and/or change the meaning of this answer 👍

xoen (approved):

PR #448, opened 2025-05-22
IR-775: Description addenda

ushkarev:


server/routes/reports/descriptionAddendum/index.ts
```diff
@@ -26,9 +27,19 @@ class AddDescriptionAddendumController extends BaseController<Values> {
     }
   }
 
+  private async lookupUsers(
+    _req: FormWizard.Request<Values>,
+    res: express.Response,
+    next: express.NextFunction,
+  ): Promise<void> {
+    const report = res.locals.report as ReportWithDetails
+    res.locals.usersLookup = await res.locals.apis.userService.getUsers(res.locals.systemToken, [report.reportedBy])
```
xoen, line 40: Potentially we could lookup other usernames too, like ones in correction requests and staff involvement?

And maybe use this middleware in other routes that needs these users' details?

(Just a thought)

ushkarev: but middleware _could_ be useful cos this creator name is needed in a few places

ushkarev: we don’t need to look up names in staff/prisoner involvements or addenda

PR #461, opened 2025-06-17
IR-1047: Remove canCreateReport permission flag

ushkarev:
…because

1) It implied that all users with the approve role could create some kind of report but they cannot unless they _also_ have the PECS role.
In reality `canCreateReport` should have been `hasReadWriteRole XOR (hasApproveRejectRole AND hasPecsRole)`.

2) It was too easy to misuse given that it did not take report location into account.

All permission flags reusing `canCreateReport` still had to recheck roles depending on report location anyway.

### DIFF
--- a/server/data/constants.ts
+++ b/server/data/constants.ts
@@ -34,6 +34,8 @@ export const roleApproveReject = 'INCIDENT_REPORTS__APPROVE' as const
  */
 export const rolePecs = 'INCIDENT_REPORTS__PECS' as const
 
+// TODO: will management reporting need a separate role?
+
 /**
  * Used by DPS/NOMIS central admin/support.
  */
--- a/server/views/pages/index.njk
+++ b/server/views/pages/index.njk
@@ -47,6 +47,7 @@
       description: "TODO",
       href: "/reports"
     }) }}
+    {# TODO: will the management reporting tile need a separate role? #}
     {{ renderDpsCard({
       heading: "Management reporting",
       description: "Run management reports for printing or exporting.",
--- a/server/routes/dpr/index.ts
+++ b/server/routes/dpr/index.ts
@@ -46,6 +46,8 @@ export function dprRouter(router: Router, services: Services): Router {
     getIncidentReportingApi(services).then(incidentReportingApi => populateRoutes(incidentReportingApi, router))
   }
 
+  // TODO: will management reporting need a separate role?
+
   router.get('/management-reporting', async (req, res) => {
     const { incidentReportingApi } = res.locals.apis
     const definitions = await populateRoutes(incidentReportingApi, router)


server/data/constants.ts
```diff
@@ -34,6 +34,8 @@ export const roleApproveReject = 'INCIDENT_REPORTS__APPROVE' as const
  */
 export const rolePecs = 'INCIDENT_REPORTS__PECS' as const
 
+// TODO: will management reporting need a separate role?
```
Mjwillis, line 37: This is a good question, my feeling is no, as each report is controlled by an existing role and the caseloads.  We might need something specific but I think users can see reports on the prisons they working in

server/views/pages/index.njk
```diff
@@ -47,6 +47,7 @@
       description: "TODO",
       href: "/reports"
     }) }}
+    {# TODO: will the management reporting tile need a separate role? #}
```
Mjwillis, line 50: Within the page there are checks for each report.

Mjwillis (approved):
LGTM

server/data/constants.ts
ushkarev: is there any report that’s open to all ppl?

server/middleware/permissions/rulesClass.ts
```diff
@@ -51,17 +52,12 @@ export class Permissions {
...
     return (
-      // has minimal roles
-      this.canCreateReport &&
       // if PECS region, can user edit PECS reports (has approver & PECS roles)?
-      ((isPecsRegionCode(code) && this.roles.has(roleApproveReject) && this.roles.has(rolePecs)) ||
-        // otherwise does user have caseload but isn't a data warden?
-        (this.caseloadIds.has(code) && !this.roles.has(roleApproveReject)))
+      (isPecsRegionCode(code) && this.roles.has(roleApproveReject) && this.roles.has(rolePecs)) ||
+      // otherwise does user have caseload but isn't a data warden?
+      (this.caseloadIds.has(code) && this.roles.has(roleReadWrite) && !this.roles.has(roleApproveReject))
```
xoen, line 60: We could give this (and probably other conditions) a name, e.g. something like `this.isReportingOfficerIn(code)`.

Similarly, we could have a `this.isDataWarden()` and also `this.canReportAndApprovePecs(code)` or something like that.

I'm just thinking aloud here, but I do wonder if potentially these could be building blocks with nice names as I've seen more or less the same conditions in a few places.

PS: Do we need to check the absence of the `roleApproveReject` role? 🤔 I know there shouldn't be overlapping _**in theory**_ but I do wonder what it adds.

xoen: But why would we want to break the experience for a user just because it has both hats? 🤔 

Somehow in my mind I see these roles as additive, e.g. you have the `readWrite` role and the prisonId is your caseload...shouldn't you then be able to create a report (don't matter if you also happen to have the `approveReject` role (even tho, granted a user should not have both in practice).

xoen: As discussed offline, the main argument to have that additional check is a failsafe in case a user is accidentally (or else) be given both the readWrite role (RO) and the approveReject role (DW).

This kinda goes against the requirement of preventing someone from being both a Reporting Officer and a Data Warden as it would mean someone could both create a report and also approve it themselves.

PR #474, opened 2025-06-30
IR-1097: Renaming config fields to match API names

xoen:
UI names have been different from the API:
- numeric identifier, e.g. `12345` for Qs/As:
  - API calls these `code`, UI called these `id` (now renamed to `code`)
- string identifier, e.g.  `IS LOCATION KNOWN` (the version of question/answer in NOMIS, can be thought as constant)
  - API calls these `question`/`response`, UI called these `code` (now renamed to `question`/`response`)
- human version of the Qs/As with improved spelling, punctuation, etc...
  -  API doesn't have this as it's in the UI report type config, UI calls this `label` e.g. `Is location known?`

The mismatch between UI and API names has been quite confusing, especially at the edge when communicating with the API. But in general it wasn't always apparent if `code` meant the number or the string. This should make everything much clearer.


Mjwillis (approved):
LGTM

PR #475, opened 2025-07-03
IR-1102: Send question/answer labels to the API

xoen:
Depends on API PR: https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/350/

### DIFF
--- a/server/data/testData/incidentReporting.ts
+++ b/server/data/testData/incidentReporting.ts
@@ -93,12 +93,12 @@ export function mockReport({
         questions: buildArray(2, questionIndex => ({
           code: (questionIndex + 1).toString(),
           question: `Historic question #${questionIndex + 1}`,
-          label: `Historic question #${questionIndex + 1}`,
+          label: `Historic question label #${questionIndex + 1}`,
           additionalInformation: '',
           responses: buildArray(2, responseIndex => ({
             code: `${responseIndex + 1}`,
             response: `Historic response #${responseIndex + 1}`,
-            label: `Historic response #${responseIndex + 1}`,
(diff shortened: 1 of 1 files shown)


integration_tests/e2e/createCompletedReport.cy.ts
```diff
@@ -415,16 +415,20 @@ context('Creating a completed draft report', () => {
 const apiQuestionResponse = (
   code: string,
   question: string,
-  response: string,
```
xoen, line 418: I moved the response after the code for consistency

server/data/testData/incidentReporting.ts
```diff
@@ -209,10 +210,12 @@ export function mockQuestion(
   return {
     code: (questionIndex + 1).toString(),
     question: `Question #${questionIndex + 1}`,
+    label: `Question #${questionIndex + 1}`,
```
jonesj2193, line 213: Same here with label instead of question? or even "Question label" ?

server/data/testData/incidentReporting.ts
```diff
@@ -94,10 +93,12 @@ export function mockReport({
         questions: buildArray(2, questionIndex => ({
           code: (questionIndex + 1).toString(),
           question: `Historic question #${questionIndex + 1}`,
+          label: `Historic question #${questionIndex + 1}`,
```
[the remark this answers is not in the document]
xoen: Not too bothered either way but I've tweaked these in this commit anyway: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/475/commits/21461102d0f5917de347d85080c5169bd47517a4

jonesj2193 (approved):

PR #479, opened 2025-07-15
IR-1134: Handle 404 from /agencies-switches

xoen:
In productions we are seeing 404 Not Found errors. These are handled ok but the interesting thing is
that this is the way that endpoing behaves when no agencies are active.

This commit handle this scenario more explicitly so that the errors are not concerning anybody and also to document this is the way the endpoint works.

### DIFF
--- a/server/data/prisonApi.test.ts
+++ b/server/data/prisonApi.test.ts
@@ -89,8 +89,23 @@ describe('prisonApi', () => {
         .matchHeader('authorization', `Bearer ${accessToken}`)
         .reply(200, expectedResponse satisfies ActiveAgency[])
 
-      const activeAgenciesMiss = await apiClient.getAgenciesSwitchedOn()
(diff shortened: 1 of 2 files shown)


server/data/prisonApi.test.ts
```diff
@@ -89,8 +89,23 @@ describe('prisonApi', () => {
         .matchHeader('authorization', `Bearer ${accessToken}`)
         .reply(200, expectedResponse satisfies ActiveAgency[])
 
-      const activeAgenciesMiss = await apiClient.getAgenciesSwitchedOn()
-      expect(activeAgenciesMiss).toEqual(expectedActiveAgencies)
+      const activeAgencies = await apiClient.getAgenciesSwitchedOn()
+      expect(activeAgencies).toEqual(expectedActiveAgencies)
+    })
```
xoen, line 94: FYI: Just renamed the variable in the other test.

PR #493, opened 2025-07-30
IR-1141 Refactor secrets configuration in values.yaml

Mjwillis:


helm_deploy/hmpps-incident-reporting/values.yaml
```diff
@@ -35,13 +35,17 @@ generic-service:
     AUDIT_SERVICE_NAME: "UNASSIGNED"
     LOAD_DPR_ON_STARTUP: "true"
 
-  envFrom:
-    - secretRef:
-        name: hmpps-incident-reporting
-    - secretRef:
-        name: hmpps-incident-reporting-application-insights
-
```
ushkarev, line 43: why not just this? since all secrets are exported without being renamed. simpler, no? less error prone

```yaml
  envFrom:
    - secretRef:
        name: hmpps-incident-reporting-session-secret
    - secretRef:
        name: hmpps-incident-reporting-auth-code
    - secretRef:
        name: hmpps-incident-reporting-client-creds
    - secretRef:
        name: hmpps-incident-reporting-application-insights
```

Mjwillis: It not the standard and I hate it too. Its not explicit - you can't see what secrets you are using.

Mjwillis: Also they don't match the new values

ushkarev: sure, the values won’t match, that’s why i didn’t include the old `hmpps-incident-reporting` k8s secret in the snippet above. but when i made it, you weren’t renaming the env vars and now you are ;)

ushkarev (approved):

PR #504, opened 2025-08-05
IR-1148: Call where an incident happened “Location”

ushkarev:


server/routes/reports/viewReport.ts
```diff
@@ -69,9 +69,9 @@ export function viewReportRouter(): Router {
       if (report.correctionRequests) {
         usernames.push(...report.correctionRequests.map(correctionRequest => correctionRequest.correctionRequestedBy))
       }
-      const [usersLookup, prisonsLookup] = await Promise.all([
+      const [usersLookup, locationLookup] = await Promise.all([
         userService.getUsers(res.locals.systemToken, usernames),
-        prisonApi.getPrisons(),
+        prisonApi.getPrisons(), // TODO: wait why is this looking up ALL prisons?
```
xoen: is this used in the filter? oh but it's the single report page so why would it need that, should it just need the report's location 🤔?

PR #523, opened 2025-08-19
IR-908: Initial PECS work

ushkarev:
- Adds some entry points to create a PECS report: but the actual journey does NOT exist yet. PECS reports cannot be created yet.
- Adds additional filtering options to the dashboard based on location

### DIFF
--- a/integration_tests/e2e/dashboard.cy.ts
+++ b/integration_tests/e2e/dashboard.cy.ts
@@ -163,6 +163,7 @@ context('Searching for a report', () => {
     dashboardPage.locationOptions.then(locationOptions => {
       expect(locationOptions).to.deep.equal([
         { label: 'All locations', value: '' },
+        { label: 'All locations active in the service', value: '.ACTIVE' }, // TODO: remove after rollout
         { label: 'All PECS regions', value: '.PECS' },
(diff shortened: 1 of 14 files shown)


integration_tests/e2e/dashboard.cy.ts
```diff
@@ -150,6 +151,25 @@ context('Searching for a report', () => {
...
+    cy.task('stubIncidentReportingApiGetReports')
+    cy.visit('/reports')
+    const dashboardPage = Page.verifyOnPage(DashboardPage)
+    dashboardPage.locationOptions.then(locationOptions => {
+      expect(locationOptions).to.deep.equal([
+        { label: 'All locations', value: '' },
+        { label: 'All PECS regions', value: '.PECS' },
+        { label: 'Moorland (HMP & YOI)', value: 'MDI' },
+        { label: 'PECS North', value: 'NORTH' },
+        { label: 'PECS South', value: 'SOUTH' },
```
xoen, line 170: Are we putting the specific PECS region at the very bottom - separating them from 'All PECS regions' - because we think filtering by specific PECS region is a relatively infrequent thing users will do?

xoen (approved):

PR #526, opened 2025-08-20
Show question & response *labels* on secret type history page

ushkarev:
… so that the human-visible content is used in preference over internal/NOMIS text (when available)

### DIFF
--- a/server/views/pages/reports/history/type.njk
+++ b/server/views/pages/reports/history/type.njk
@@ -41,7 +41,7 @@
 
                   {% if inset %}
                     <div class="app-response--inset">
-                      {{ response.response }}
+                      {{ response.label }}
                       {% if response.responseDate %}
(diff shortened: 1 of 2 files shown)


server/views/pages/reports/history/type.njk
```diff
@@ -41,7 +41,7 @@
 
                   {% if inset %}
                     <div class="app-response--inset">
-                      {{ response.response }}
+                      {{ response.label or response.response }}
```
xoen, line 44: aren't the labels always there? 🤔

ushkarev: yeah you’re right. db has no `''` nor `null` so i have redone it 0d203e8c72c79e7ec0799efb1eb28ed4e2f91370

xoen (approved):

PR #533, opened 2025-08-29
Banner changes

jonesj2193:


server/middleware/permissions/statusTransitions.ts
```diff
@@ -323,10 +323,8 @@ export const pecsReportTransitions: Transitions = {
         newStatus: 'CLOSED',
         label: 'Close report',
         mustBeValid: true,
-        successBanner: 'Incident report $reportReference has been marked as closed',
+        successBanner: 'Incident report $reportReference is now closed',
       },
-      // it would be logical to also allow MARK_DUPLICATE & MARK_NOT_REPORTABLE,
-      // but it’s been decided that those should only be possible after reopening
```
ushkarev, line 329: I had intentionally put these notes in to keep a record of an alternative status chart which the stakeholders had signed off, but which disappeared during design

PR #539, opened 2025-09-04
Adding last updated field to top of view report page

jonesj2193:
Adding the entry at the top of the view report page to show who last made a change to the report and when. It was decided by Steve to leave the time off for now to keep in line with the existing reported by entry [IR-1175]


Mjwillis (approved):
LGTM

PR #544, opened 2025-09-08
IR-1174 Incident type audit history

jonesj2193:
Adding the incident type audit history to the view report page. 

work includes:

- Adding updates to incident reporting api calls to include new incident type history
- Updating template for view report page to display incident type history if one does exist
- Unit and e2e tests updated to check rendering as expected


Mjwillis (approved):
lgtm

PR #560, opened 2025-09-25
IR-1299: Incident type history shows most recent changes on top

xoen:
The report's type history format is a bit awkard to present. This has made the template a bit more complicated than it probably should be. It also makes it tricky to change the order of the changes displayed.

I've moved the logic that parse this format into an helper function used by the request handler so that the template can _almost_ display this list as-is without having to massage it.

When a report type change as follows:

```
Type A @T1 ==> Type B @ T2 ==> Type C @ T3
```

(where the report type at creation was Type A and its current type is Type C - user omitted for clarity)

The `report.incidentTypeHistory` is as follows:

- Type A @ T2
- Type B @ T3

Note how:
1. The creation time/author is not in this list (they're in the report's reportedAt/By field)
2. The time T2 is not the creation type but rather when the report type changed **from** Type A to Type B
3. The final/current type is not in this list (it's the `report.type` field)

What we really want to display is something like:
- Created as Type A @ T1 by U1
- Changed to Type B @ T2 by U2
- Changed to Type C @ T3 by U3

This is what the template was doing, but now this logis is moved in the `cleanTypeHistory()` function which returns an array where the first element is the creation type/time/author, and a subsequent element for each of the changes to the given type/time/author (including the last change to the current report's type).

### DIFF
--- a/server/routes/reports/viewReport.test.ts
+++ b/server/routes/reports/viewReport.test.ts
@@ -358,11 +358,23 @@ describe('View report page', () => {
         .get(viewReportUrl)
         .expect('Content-Type', /html/)
         .expect(res => {
+          // Check content
           expect(res.text).toContain('Incident created as Assault by John Smith on 5 December 2023 at 12:34')
           expect(res.text).toContain('Incident updated to Drone sighting by John Smith on 5 December 2023 at 12:34')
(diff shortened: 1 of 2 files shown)


jonesj2193 (approved):
LGTM

PR #565, opened 2025-10-02
IR-1215 removals filter

jonesj2193:
Adding new filter to enable DWs to filter for removal requests. 

Work done:

- Updated API call and associated types to filter for removal requests via `lastUserAction` parameter
- Updated basic report type to include new variable `lastUserAction`
- Added filter with validation on to the dashboard. 

No tests have been added for this yet. 

No unit tests have been broken, integration tests are now broken.

### DIFF
--- a/server/data/incidentReportingApi.ts
+++ b/server/data/incidentReportingApi.ts
@@ -107,7 +107,7 @@ export type GetReportsParams = {
   reportedByUsername: string
   involvingStaffUsername: string
   involvingPrisonerNumber: string
-  latestUserAction: ApiUserAction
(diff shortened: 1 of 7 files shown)


server/data/incidentReportingApi.ts
```diff
@@ -107,7 +107,7 @@ export type GetReportsParams = {
   reportedByUsername: string
   involvingStaffUsername: string
   involvingPrisonerNumber: string
-  latestUserAction: ApiUserAction
+  userAction: ApiUserAction | ApiUserAction[]
```
xoen, line 110: Isn't the API just accepting an array? 🤔

PR #794, opened 2026-04-10
IR-1574: Use modern `module`/`moduleResolution`

xoen:
- Updated `tsconfig.json` files to use modern `module`/`moduleResolution`:
  - use `module` setting coming from `@tsconfig/node24/tsconfig.json`, this means using `NodeNext` instead of `CommonJS` ([see `module` documentation](https://www.typescriptlang.org/tsconfig/#module))
   - use `moduleResolution` coming from `@tsconfig/node24/tsconfig.json`, this means using `Node16` instead of `Node` ([see `moduleResolution` documentation](https://www.typescriptlang.org/tsconfig/#moduleResolution))
   - these two combined "picks the right algorithm for each resolution based on whether Node.js will see an `import` or `require` in the output JavaScript code."
 - Fix dynamic `import` (used to load Incident Types configurations based depending on report) which now works the same way regardless of when we run the transpiled JS or the TS (during tests)
   - NOTE: For the dynamic `import` to work jest needs support for ESM modules (ECMAScript Modules) via the `--experimental-vm-modules` flag. This is what is documented in [the `jest` docs](https://jestjs.io/docs/next/ecmascript-modules).
- updated version of JS integration tests JS is compiled to `es2024` (from `es2022`), for the integration tests it shouldn't cause issues
- removed `"esModuleInterop": true` and `"skipLibCheck": true` as these settings are the same in the extended `@tsconfig/node24/tsconfig.json` file so we can remove without having any impact - but makes the `tsconfig.json` files simpler
- finally, I noticed that there were a number of _extraneous_ packages, packages that shouldn't be needed as they're not a direct or transient dependency. So I've run `npm prune` to remove them from the `package-lock.json` (npm _should_ remove packages as it goes during `npm install`, etc but for whatever reason it didn't with these).

Also cherry-picked [this renovate config change](https://github.com/ministryofjustice/hmpps-template-typescript/commit/bfd4e449c4a55b070557fe80967038eebd365687) from TS template.


assets/js/tsconfig.json
```diff
@@ -2,12 +2,9 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "module": "commonjs",
     "target": "es2020",
     "noEmit": true,
     "lib": ["dom", "dom.iterable", "es2020"],
-    "esModuleInterop": true,
```
xoen, line 9: These have no effect in practice as `@tsconfig/node24/tsconfig.json` - the file we extend from in the main `tsconfig.json` - has the same values.

integration_tests/tsconfig.json
```diff
@@ -2,16 +2,13 @@
...
-    "lib": ["dom", "es2022"],
+    "lib": ["dom", "es2024"],
     "types": ["cypress", "express", "express-session"],
-    "esModuleInterop": true,
-    "skipLibCheck": true,
     "sourceMap": false,
     "typeRoots": ["../node_modules/@types", "../node_modules", "../server/@types"]
   },
   "exclude": [],
-  "include": ["../server/@types/**/*.d.ts", "**/*.ts"],
```
xoen, line 16: Removed trailing comma (invalid in JSON in theory) to make precommit hook happy.

server/reportConfiguration/types/index.ts
```diff
@@ -13,11 +13,5 @@ export function getIncidentTypeConfiguration(type: string): Promise<IncidentType
   const ext = __filename.endsWith('.js') ? 'js' : 'ts'
   const configPath = path.resolve(__dirname, `./${type}.${ext}`)
 
-  return import(configPath).then(module => {
```
xoen, line 16: Before the change to `module`/`moduleResolution` the structure of the module yielded changed slightly depending on whether we were running the TS (during tests) or the transpiled JS.

Now they both have the same structure.

tsconfig.json
```diff
@@ -4,8 +4,6 @@
   "compileOnSave": true,
   "compilerOptions": {
     "strict": false,
-    "module": "CommonJS",
```
xoen, line 7: Using values coming from [`@tsconfig/node24/tsconfig.json`](https://www.npmjs.com/package/@tsconfig/node24):

```JSON
"module": "nodenext",
"target": "es2024",
```

Mjwillis (approved):
LGTM

PR #795, opened 2026-04-14
IR-1576 Remove Management Reporting feature and related configurations

Mjwillis:
- Removed DPR-related environment variables from multiple configuration files such as values.yaml, .env.sample, and others.
- Deleted DPR-related UI elements, including tiles, routes, and templates (e.g., managementReporting.njk).
- Eliminated DPR-related test code, removing references from integration tests and Cypress tests.
- Updated backend code, removing functions, constants, and imports associated with DPR functionality.
- Removed DPR-specific documentation mention from the technical overview.


xoen (approved):
🙌

PR #798, opened 2026-04-15
IR-1577: bespoke incident summary hint text

jonesj2193:
Added a new function into the `incidentDateAndTime` controller to overwrite the hint text on the incidentDate field if required. 
The wording of the hint text, for which type and field are defined in a new config object. Typing on this object could be made stricter (i.e. the keys are only set to be a string and not `incidentDate |` etc. 
Unit tests added to check correct hint text appears in the correct incident types.


Mjwillis (approved):
Perfect approach!

server/reportConfiguration/constants/types.ts
```diff
@@ -336,6 +336,13 @@ export const typesDescriptions: Record<Type, string> = {
   TOOL_LOSS_1: 'Tool or implement loss',
 }
 
+/** Adding bespoke hint text for incident types */
+export const typeIncidentDateHints: { [K in Type]?: Record<string, string> } = {
```
xoen, line 340: the name of this `const` implies it would just contains hints for the incident date field. Does the value need to be a dictionary? Couldn't it be simpler and just be the hint `string`, e.g. something like:

```TypeScript
export const typeIncidentDateHints: { [K in Type]?: Record<string, string> } = {
  RELEASE_IN_ERROR_1: 'This should be the date the person was released - for example, 17/5/2024',
}
```

server/routes/reports/details/incidentDateAndTimeController.ts
```diff
@@ -10,16 +10,37 @@ import {
...
+export abstract class BaseIncidentDateAndTimeController<
+  V extends IncidentDateAndTimeControllerValues,
+> extends BaseController<V, IncidentDateAndTimeFieldNames> {
+  middlewareLocals(): void {
+    this.use(this.customiseFields)
+    super.middlewareLocals()
+  }
+
+  /** Rewrite hint text for incident date if required */
+  private customiseFields(
```
xoen, line 30: Maybe rename to something like `customiseIncidentDateHint()` (`overrideIncidentDateHint()`?) given we are not overriding anything else (and don't think we're planning to do anyway)

server/reportConfiguration/constants/types.ts
jonesj2193: Oh yeah good point, that was an older version I had before I opened it up to be more generic.

PR #802, opened 2026-04-18
Upgrade TypeScript to v6

xoen:
TypeScript 6 has some breaking changes, mainly the `"types"` options now defaults to `[]`, which means type definitions are no longer automatically used when found in the `"typeRoots"` paths.

Also:
- improved/simplified `tsconfig.json` files (e.g. removed redundant options already set in the parent(s) config(s))
- enabled strict type checking on frontend TypeScript
- cherry-picked some of the changes from TS template
- removed type definitions that are not provided by the `@ministryofjustice/frontend-types` package
- upgraded to Cypress `15.14.2` in order to successfully run integration tests
- disabled deprecated and potentially unsafe `Cypress.env()`
- removed dependency to `uuid` and replace this with standard JS' `crypto.randomUUID()`


Mjwillis (approved):
lgtm

PR #804, opened 2026-04-20
IR-1582: Hyphen Self-harm consistently

xoen:
- Hyphenate 'Self-harm' name when creating a report or when filtering etc (note the types and type family constants need to change in the API as, [there is a PR](https://github.com/ministryofjustice/hmpps-incident-reporting-api/pull/554) for that)
- Hyphenate 'Self-harm' in questions and answers labels (both in the Self-harm type but also in Disorder which mention self-harm in an answer)
- Added hint on Self-harm question about cuts
- Fix regression on Key or lock compromise 'Heathcare centre' typo (see commit for context)
- Re-generate graphs:
  - some were slightly out of date after the Yes/No reordering
  - generate graphs for new Business Continuity types


jonesj2193 (approved):
LGTM

Mjwillis (approved):
LGTM

PR #807, opened 2026-04-22
IR-1545: DPR signpost

xoen:
Added homepage signpost to point users toward the DPR platform where they can now run reports.

Screenshot 2026-04-22 at 15 01 21

### DIFF
--- a/server/views/pages/index.njk
+++ b/server/views/pages/index.njk
@@ -65,6 +65,14 @@
 
   {% endcall %}
 
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <p class="dpr-signpost">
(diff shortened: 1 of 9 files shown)


server/views/pages/index.njk
```diff
@@ -65,6 +65,14 @@
 
   {% endcall %}
 
+  <div class="govuk-grid-row">
```
jonesj2193, line 68: Do we want to put an if condition around this as with the sharepoint url for consistency? Just to ensure that there is a working url there and prevent users clicking on a broken link/one that isn't correctly configured. 
Happy to leave this, just thought I would ask the question.

xoen: I think it's overkill as the URL is always set in all environment via the Helm's `values-*.yaml` files unlike the Sharepoint URL.

Also, the `dprUrl` is read using `requiredInProduction` so it wouldn't even reach this point anyway if the URL is not present so it would just be cluttering the template with no added benefit.

jonesj2193 (approved):

PR #817, opened 2026-05-01
`prisonerLocation()` refactoring

xoen:
As a result of the upgrade to Cypress 15 I discovered some weirdness related to the logic that displays the prisoner location. Specifically the `prisonerLocation()` utility function wasn't really used in its entirety, this was copied from Non-associations and/or Incentives at some point and was still displaying the cell location (which we don't show in this service).

Furthermore the nunjucks template was almost duplicating this logic (but with slight variations, e.g. showing 'N/A' instead of 'Transfer').

I've retaining the existing behaviour, e.g. this is a refactoring and _**it should not change**_ what the user see.
But I've simplified things so that there is less confusion in the future:
- removed nunjucks logic/moved it into the `prisonerLocation()` utility function (which now has the same logic as the template)
- don't add `isBeingTransferred` and `isOutside` as nunjucks filters as these are no longer used directly
  - also, these are not exported in the module where they live as they're an internal implementation detail
- the `isInPrison` function was removed as it was not used (I'm assuming it was just added "just-in-case" but YAGNI)
- simplified this function's tests as a result


server/views/pages/prisoners/search.njk
```diff
@@ -168,14 +168,6 @@
           <img src="/prisoner/{{ result.prisonerNumber }}/photo.jpeg" alt="Photo of {{ prisonerName | safe }}" class="app-prisoner-photo" />
         {% endset %}
 
-        {% if result | isBeingTransferred %}
```
xoen, line 171: Logic is now moved in `prisonerLocation()`

server/utils/utils.ts
```diff
@@ -106,19 +103,6 @@ export function datesAsStrings<T>(obj: T): DatesAsStrings<T> {
   return Object.fromEntries(Object.entries(obj).map(([property, value]) => [property, datesAsStrings(value)]))
 }
 
-/**
- * Display location of a prisoner in prison, during transfer and outside/released
- */
-export const prisonerLocation = (prisoner: OffenderSearchResult): string => {
```
xoen, line 112: This was moved in `prisonerLocationUtils` module and tweaked to behave like the old version of the nunjucks template logic.

server/utils/utils.test.ts
```diff
@@ -63,86 +60,6 @@ describe('display of prisoner names', () => {
   })
 })
 
-describe('prisoners’ locations', () => {
```
xoen, line 66: Tests also moved.

server/utils/prisonerLocationUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+  OffenderSearchResultTransfer,
+} from '../data/offenderSearchApi'
+
+function isBeingTransferred(prisoner: OffenderSearchResult): prisoner is OffenderSearchResultTransfer {
+  return prisoner.prisonId === transferPrisonId
+}
+
+function isOutside(prisoner: OffenderSearchResult): prisoner is OffenderSearchResultOut {
+  return prisoner.prisonId === outsidePrisonId
+}
```
xoen, line 14: These 2 functions are now private as they're only used inside `prisonerLocation()` and we don't need to add them to the public interface of this module.

server/utils/prisonerLocationUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Display location of a prisoner in prison, during transfer and outside/released
+ */
+export function prisonerLocation(prisoner: OffenderSearchResult): string {
+  if (isBeingTransferred(prisoner)) {
+    return 'N/A'
+  }
+  if (isOutside(prisoner)) {
+    return prisoner.locationDescription || 'Outside'
+  }
+  return prisoner.prisonName || 'Not known'
```
xoen, line 26: This is now doing the [same thing that the nunjucks template was doing before](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/32cac3deea62f3a93f2e7afa4297ef6e57b8c66c/server/views/pages/prisoners/search.njk#L171-L177) (e.g. no change in what we display to users)

server/utils/nunjucksSetup.ts
```diff
@@ -106,9 +98,6 @@ export default function nunjucksSetup(app: express.Express): void {
 
   // prisoner utils
   njkEnv.addFilter('prisonerLocation', prisonerLocation)
-  njkEnv.addFilter('isBeingTransferred', isBeingTransferred)
-  njkEnv.addFilter('isOutside', isOutside)
-  njkEnv.addFilter('isInPrison', isInPrison)
```
xoen, line 111: Don't need to add these functions as nunjucks filters as they're not used in any template (YAGNI)

server/utils/prisonerLocationUtils.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  describe('for people being tranferred', () => {
+    it(`returns 'N/A'`, () => {
+      expect(prisonerLocation(donald)).toEqual('N/A')
+    })
+
+    it(`without a location description, also returns 'N/A'`, () => {
+      const prisoner = {
+        ...donald,
+        // @ts-expect-error - invalid input for testing purpose
```
xoen, line 19: This directive is to tell TS that we know this value is not proper and we expect a type error, but that's by design (to test how the function works with invalid input).

NOTE: We may have to tweak the test further if/when we enable `strict: true` as I couldn't find a version of this that would make TS happy now and with `strict: true`, that is without using type assertions (`as T`) which are not good!

Mjwillis (approved):
LGTM

PR #826, opened 2026-05-08
IR-1605: Removed Admin screen

xoen:
As NOMIS screens are off for all prisons we don't need these admin screens anymore.


server/data/prisonApi.ts
```diff
@@ -222,45 +222,6 @@ export class PrisonApi extends RestClient {
...
-    } catch (error) {
-      if (error?.responseStatus === 404) {
-        // endpoint returns 404s when agency is not active
-        return false
-      }
-      throw error
-    }
-  }
-
-  async activatePrison(prisonId: string): Promise<ActiveAgency> {
```
xoen, line 244: FYI: The `ActiveAgency` type is still used and I haven't removed it as part of this PR, it'll remove that in a separate PR.

server/data/prisonApi.ts
```diff
@@ -222,94 +201,6 @@ export class PrisonApi extends RestClient {
     }
   }
 
-  async isPrisonActive(prisonId: string): Promise<boolean> {
```
Mjwillis, line 225: Isn't isActivePrison used to determine whether this incident can be edited?

Mjwillis (commented):
Comment about the use of activeprison?

server/data/prisonApi.ts
xoen: I don't see any other use of this function, I think it's using the `activeAgencies` module/field which is updated by a middleware to cache the list of active agencies.

But I haven't removed that just yet/as part of this as it's a bit more involved and I need to check how things are wired together and it's a bit more involved (hence why I'm leaving it out of this PR that is specifically to remove the Admin screens).

Mjwillis (approved):
LGTM

PR #828, opened 2026-05-11
IR-1609: Remove 'All active locations in Digital ...'

xoen:
Removed the option to filter reports in locations currently active in DPS ('All active locations in Digital Prison Services (DPS)'). All prisons are usingDPS not so this is the same as the "All locations" option.


Mjwillis (dismissed):

server/middleware/permissions/rulesClass.ts
```diff
@@ -143,7 +115,7 @@ export class Permissions {
 
     const allowedActions = new Set<UserAction>(['VIEW'])
 
-    if ((where === 'dps' && !locationIsActive) || (where === 'nomis' && locationIsActive)) {
```
xoen, line 146: If you work this out given `locationIsActive` would be now always `true` this `if` condition becomes just `where === 'nomis'`:

```TypeScript
(where === 'dps' && !locationIsActive) || (where === 'nomis' && locationIsActive)

// equivealent to...
(where === 'dps' && !true) || (where === 'nomis' && true)

// equivealent to...
(where === 'dps' && false) || (where === 'nomis' && true)

// equivealent to...
(false) || (where === 'nomis')

// equivealent to...
where === 'nomis'
```

server/middleware/permissions/rulesClass.ts
```diff
@@ -143,7 +115,7 @@ export class Permissions {
...
+    if (where === 'nomis') {
       // only modifiable in nomis
```
Mjwillis: Won't this always be false as nothing is only modifiable in nomis

server/middleware/permissions/rulesClass.ts
Mjwillis: `where` will never be set to `nomis` I think

Mjwillis (commented):
Few comments

server/middleware/permissions/rulesClass.ts
xoen: I think you may be right, but I need to double check properly.

(Now that we removed a lot of code this may be apparent but I tried to refactor things in steps rather than assume things)

server/middleware/permissions/rulesClass.test.ts
```diff
@@ -154,18 +154,14 @@ describe('Permissions class', () => {
       permissions: Permissions
       userActions: UserAction[]
       all: 'granted' | 'denied'
-      onlyInNomis?: true
     }
 
-    function expectActionsOnReports(
-      reports: ReportBasic[],
-      { permissions, userActions, all, onlyInNomis }: Expectation,
```
xoen, line 162: removed `onlyInNomis` it was no longer passed and as a result it was always destructured as `undefined`.

server/middleware/permissions/rulesClass.test.ts
```diff
@@ -154,18 +154,14 @@ describe('Permissions class', () => {
...
-    ) {
+    function expectActionsOnReports(reports: ReportBasic[], { permissions, userActions, all }: Expectation) {
       for (const report of reports) {
-        const allowedActions = permissions.allowedActionsOnReport(report, onlyInNomis ? 'nomis' : 'dps')
+        const allowedActions = permissions.allowedActionsOnReport(report)
         expect(userActions.every(userAction => allowedActions.has(userAction))).toBe(all === granted)
 
-        if (!onlyInNomis && all === 'granted') {
```
xoen, line 168: as `onlyInNomis` is now always `undefined`, `!onlyInNomis` is always `true` and it can be removed from the `if` condition.

server/middleware/permissions/rulesClass.ts
```diff
@@ -94,16 +94,10 @@ export class Permissions {
...
    * Returns the set of user actions allowed on this report.
-   * Actions that change a report are permitted either only on DPS or only in NOMIS.
-   * Set `where` to “nomis” to get actions that would have been allowed if the location were active in DPS –
-   * this indicates that users should go to NOMIS to complete their intended work.
    *
    * NB: this does NOT perform report validity checks!
    */
-  allowedActionsOnReport(
-    reportLike: Pick<ReportBasic, 'status' | 'location'>,
-    where: 'dps' | 'nomis' = 'dps',
```
xoen, line 105: `where` parameter no longer passed (with the exception of that test updated in this commit)

server/middleware/permissions/rulesClass.ts
```diff
@@ -115,11 +109,6 @@ export class Permissions {
 
     const allowedActions = new Set<UserAction>(['VIEW'])
 
-    if (where === 'nomis') {
```
xoen, line 118: `where` never passed so it was defaulting to `'dps'`, meaning this if condition is now always `false`.

server/middleware/permissions/rulesClass.ts
xoen: Yes! Well spotted. I've removed the `where` parameter and updated the test in this commit: https://github.com/ministryofjustice/hmpps-incident-reporting/pull/828/commits/2c6ae917cefd3064abb5fa6500f50e0928e371f7

Mjwillis (approved):
LGTM

PR #837, opened 2026-05-14
IR-1635: Use system token for Prison API prisoner photo requests

Mjwillis:
# IR-1635: Use client credentials token for Prison API prisoner photo requests

## Summary

Fixes 403 errors when fetching prisoner photos for users who have global search access but do not have the specific establishment in their active caseload.

Prison API's `/api/bookings/offenderNo/{prisonerNumber}/image/data` endpoint enforces caseload-level access when called with a user's delegated OAuth token. This meant that users with global search permissions — who are legitimately allowed to view prisoners across establishments — would receive a 403 when viewing an incident involving a prisoner not in their active caseload.

The fix is to call Prison API with a **client credentials (system) token** for image fetches. The decision of whether to display the photo remains with the frontend application.

### `server/routes/index.ts`

The prisoner photo route (`GET /prisoner/:prisonerNumber/photo.jpeg`) was previously constructing a `PrisonApi` instance using the **user's delegated OAuth token**:

```ts
// Before
const { user } = res.locals
const prisonApi = new PrisonApi(user.token)
const photoData = await prisonApi.getPhoto(prisonerNumber)
```

It now explicitly fetches a **client credentials token** via `hmppsAuthClient.getToken(user.username)` and constructs a fresh `PrisonApi` instance with it:

```ts
// After
const { user } = res.locals
const systemToken = await hmppsAuthClient.getToken(user.username)
const prisonApi = new PrisonApi(systemToken)
const photoData = await prisonApi.getPhoto(prisonerNumber)
```

`hmppsAuthClient` is destructured from `services` at router initialisation time and is the same auth client used by `setSystemToken` middleware throughout the application. Using `getToken(username)` retrieves a cached client credentials token scoped to the user's session — it is not the user's own token.

This pattern is consistent with other places in the codebase that construct `PrisonApi` outside of `res.locals.apis` (e.g. `server/middleware/updateActiveAgencies.ts`).

### `server/data/prisonApi.test.ts`

Renamed the local test variable `accessToken` → `systemToken` across all test cases. This is a cosmetic change only — no test logic was altered — but it accurately reflects the token type that `PrisonApi` should be constructed with, removing a misleading name that implied a user access token.

## Why not use `res.locals.apis.prisonApi`?

An intermediate approach using the pre-built `res.locals.apis.prisonApi` instance (populated by `setApis` middleware) was explored but not taken forward. The explicit `hmppsAuthClient.getToken` approach was preferred because:

- It makes the token type unambiguous at the call site — a reader does not need to trace through middleware setup to understand which token is in use.
- It is self-contained within the route handler, with no dependency on the `setApis` middleware having run and populated `res.locals.apis` correctly.
- It is consistent with how other background/auxiliary operations (e.g. `updateActiveAgencies`) obtain a system token directly from `hmppsAuthClient`.

## Testing

- All existing `prisonApi.test.ts` tests continue to pass (only variable name updated).
- Manually verified that the photo endpoint no longer returns 403 for users with global search access viewing prisoners outside their active caseload.

## Risk

Low. The change targets a single auxiliary route handler. The token type (`hmppsAuthClient.getToken(username)`) is the same client credentials token used by other system-to-service calls throughout the application. No new behaviour is introduced for users within their caseload.

### DIFF
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -60,8 +60,7 @@ export default function routes(services: Services): Router {
   router.get('/prisoner/:prisonerNumber/photo.jpeg', async (req, res) => {
     const { prisonerNumber } = req.params
 
-    const { systemToken } = res.locals
-    const prisonApi = new PrisonApi(systemToken)
+    const { prisonApi } = res.locals.apis
     const photoData = await prisonApi.getPhoto(prisonerNumber)
 
     const oneDay = 86400 as const
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -58,11 +58,9 @@ export default function routes(services: Services): Router {
 
   // Auxiliary routes
   router.get('/prisoner/:prisonerNumber/photo.jpeg', async (req, res) => {
-    const { user } = res.locals
     const { prisonerNumber } = req.params
 
-    const prisonApi = new PrisonApi(user.token)
-    const photoData = await prisonApi.getPhoto(prisonerNumber)
+    const photoData = await res.locals.apis.prisonApi.getPhoto(prisonerNumber)
 
     const oneDay = 86400 as const
     res.setHeader('Cache-Control', `private, max-age=${oneDay}`)
--- a/server/routes/index.ts
+++ b/server/routes/index.ts
@@ -16,7 +16,6 @@ import { editReportRouter } from './reports/editReportRouter'
 import { reopenRouter } from './reports/actions/reopen'
 import { requestRemovalRouter } from './reports/actions/requestRemoval'
 import dashboard from './dashboard'
-import { PrisonApi } from '../data/prisonApi'
 
 export default function routes(services: Services): Router {
   const router = Router()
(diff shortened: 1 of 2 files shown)


server/routes/index.ts
```diff
@@ -58,10 +58,10 @@ export default function routes(services: Services): Router {
 
   // Auxiliary routes
   router.get('/prisoner/:prisonerNumber/photo.jpeg', async (req, res) => {
-    const { user } = res.locals
     const { prisonerNumber } = req.params
 
-    const prisonApi = new PrisonApi(user.token)
+    const { systemToken } = res.locals
+    const prisonApi = new PrisonApi(systemToken)
```
xoen: Now that you're using the system token, you could probably replace this with something simpler like:

```TypeScript
    const { prisonerNumber } = req.params
    const { prisonApi } = res.locals.apis

    const photoData = await prisonApi.getPhoto(prisonerNumber)
```

(but worth double checking if I'm missing something)

Mjwillis: Updated

xoen (dismissed):

xoen (approved):

PR #851, opened 2026-05-26
IR-1647: Allow incident type to override 'Date of incident' field label & R-1648: Hide time field for incident types where time is not relevant 

Mjwillis:
## What

Introduces a configurable label override for the `incidentDate` field so that individual incident types can replace the default "Date of incident" heading with something more semantically accurate.

Screenshot 2026-05-27 at 11 59 16

## Why

For some incident types the hint text was already customised to describe a different date (e.g. the intended release date for unlawful detention, or the actual release date for release in error), but the field label remained "Date of incident". This was misleading — the label is the prominent heading the user reads first, so it should match the meaning of the date being collected.

## How

A new `incidentTypeLabels` map has been added to `server/reportConfiguration/constants/types.ts`, parallel to the existing `incidentTypeHints` map. It follows the same `{ [K in Type]?: Record<string, string> }` shape.

The middleware method `customiseIncidentDateHint` in `BaseIncidentDateAndTimeController` has been renamed to `customiseIncidentDateFields` and extended to apply any label override from `incidentTypeLabels` via `req.form.options.fields.incidentDate.label`. This works automatically for both the create and edit flows since both use the same base controller.

Initial overrides:
- `RELEASE_IN_ERROR_1` → **"Date of release"**
- `UNLAWFUL_DETENTION_1` → **"Date meant for release"**

Adding further overrides in future only requires a new entry in `incidentTypeLabels` — no controller changes needed.

## Tests

Added to `updateIncidentDateAndTime.test.ts`:
- Parameterised test confirming regular incident types still show the default "Date of incident" label
- Test confirming `RELEASE_IN_ERROR_1` shows "Date of release" and not the default
- Test confirming `UNLAWFUL_DETENTION_1` shows "Date meant for release" and not the default

## Test plan
- [ ] Run unit tests: `npm test`
- [ ] Run type-check: `npm run typecheck`
- [ ] Manually verify the label change on the create report flow for UNLAWFUL_DETENTION_1 and RELEASE_IN_ERROR_1
- [ ] Verify other incident types still show "Date of incident"

## Summary

- Adds `incidentTypeDefaultTimes` config map in `types.ts` (same pattern as existing `incidentTypeHints`). When a type has an entry, its time-of-incident field is hidden from the user and the configured value is submitted to the API automatically.
- `UNLAWFUL_DETENTION_1` defaults to `00:00` (time of occurrence is not meaningful for this type).
- Removed the `incidentTime` hint for `UNLAWFUL_DETENTION_1` since the field is no longer shown.
- Controller changes in `BaseIncidentDateAndTimeController` cover `middlewareLocals` (hides field, sets default in `res.locals`), `getValues`, `process`, and `validate` (skips "time must be in the past" error for type-level defaults).

## Test plan

- [ ] New `describe` block in `updateIncidentDateAndTime.test.ts` verifies hidden field rendering, API submission with default time, and exemption from the future-time error
- [ ] `afterAll` reset in the new describe block prevents `reportBasic.type` mutation from bleeding into subsequent tests
- [ ] All 2595 unit tests pass (`npm test`)
- [ ] Manually verify: load a report with type `UNLAWFUL_DETENTION_1` and navigate to update-date-and-time — time field should be absent, hidden input `name="incidentTime" value="00:00"` should be present, and submitting should call the API with `00:00` as the time component

### DIFF
--- a/server/reportConfiguration/constants/types.ts
+++ b/server/reportConfiguration/constants/types.ts
@@ -359,7 +359,7 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
  * When set, the time field is hidden and this value is submitted to the API automatically.
  */
 export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
-  UNLAWFUL_DETENTION_1: '23:59',
+  UNLAWFUL_DETENTION_1: '00:00',
 }
 
 /**
--- a/server/reportConfiguration/constants/types.ts
+++ b/server/reportConfiguration/constants/types.ts
@@ -351,8 +351,15 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
   },
   UNLAWFUL_DETENTION_1: {
     incidentDate: 'The date the person was meant to be released - for example, 17/5/2024',
-    incidentTime: "For example, 09 08. If you're unsure, enter 23 59",
   },
+}
+
+/**
+ * Default times (HH:MM) for incident types where time input is not shown to the user.
+ * When set, the time field is hidden and this value is submitted to the API automatically.
+ */
+export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
+  UNLAWFUL_DETENTION_1: '23:59',
 }
 
 /**
--- a/server/reportConfiguration/constants/types.ts
+++ b/server/reportConfiguration/constants/types.ts
@@ -355,11 +355,12 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
 }
 
 /**
- * Default times (HH:MM) for incident types where time input is not shown to the user.
- * When set, the time field is hidden and this value is submitted to the API automatically.
+ * Override map for incident types where the time of incident is irrelevant.
+ * When set to false, the time field is hidden and 00:00 is submitted to the API automatically.
+ * Defaults to true (time required) for all types not listed here.
  */
-export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
-  UNLAWFUL_DETENTION_1: '00:00',
+export const incidentTypeRequiresTime: Partial<Record<Type, boolean>> = {
+  UNLAWFUL_DETENTION_1: false,
 }
 
 /**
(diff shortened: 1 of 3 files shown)


server/reportConfiguration/constants/types.ts
```diff
@@ -350,11 +350,21 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
...
   },
   UNLAWFUL_DETENTION_1: {
-    incidentDate: 'The date the person was meant to be released - for example, 17/5/2024',
+    incidentDate: 'For example, 17/5/2024',
     incidentTime: "For example, 09 08. If you're unsure, enter 23 59",
   },
 }
 
+/** Overrides for form field labels for incident types */
+export const incidentTypeLabels: { [K in Type]?: Record<string, string> } = {
```
xoen, line 367: I'd call this `incidentTypeDateLabels` instead.

xoen (changes requested):
Left a minor comment but otherwise look good.

xoen (dismissed):
Diff looks ok but some tests are failing, not sure why but worth double checking (e.g. is state/labels leaking between requests/types? 🤔)

server/routes/reports/details/updateIncidentDateAndTime.test.ts
```diff
@@ -101,6 +101,96 @@ describe('Updating report incident date and time', () => {
...
+      })
+  })
+
+  it('should show bespoke label for release in error', () => {
+    reportBasic.type = 'RELEASE_IN_ERROR_1' as Type
+    return agent
+      .get(updateIncidentDateAndTimeUrl)
+      .expect(200)
+      .expect(res => {
+        expect(res.text).toContain('Date of release')
```
xoen: This probably need to change to match the new label (currently `Date the person was released`)

server/routes/reports/details/updateIncidentDateAndTime.test.ts
```diff
@@ -101,6 +101,96 @@ describe('Updating report incident date and time', () => {
...
+      })
+  })
+
+  it('should show bespoke label for unlawful detention', () => {
+    reportBasic.type = 'UNLAWFUL_DETENTION_1' as Type
+    return agent
+      .get(updateIncidentDateAndTimeUrl)
+      .expect(200)
+      .expect(res => {
+        expect(res.text).toContain('Date meant for release')
```
xoen: Probably need to match custom label copy (currently `Date the person should have been released`)

xoen (approved):

PR #852, opened 2026-05-26
IR-1648: Hide time field for incident types where time is not relevant

Mjwillis:
## Summary

- Adds `incidentTypeDefaultTimes` config map in `types.ts` (same pattern as existing `incidentTypeHints`). When a type has an entry, its time-of-incident field is hidden from the user and the configured value is submitted to the API automatically.
- `UNLAWFUL_DETENTION_1` defaults to `00:00` (time of occurrence is not meaningful for this type).
- Removed the `incidentTime` hint for `UNLAWFUL_DETENTION_1` since the field is no longer shown.
- Controller changes in `BaseIncidentDateAndTimeController` cover `middlewareLocals` (hides field, sets default in `res.locals`), `getValues`, `process`, and `validate` (skips "time must be in the past" error for type-level defaults).

Screenshot 2026-05-27 at 09 20 11

## Test plan

- [ ] New `describe` block in `updateIncidentDateAndTime.test.ts` verifies hidden field rendering, API submission with default time, and exemption from the future-time error
- [ ] `afterAll` reset in the new describe block prevents `reportBasic.type` mutation from bleeding into subsequent tests
- [ ] All 2595 unit tests pass (`npm test`)
- [ ] Manually verify: load a report with type `UNLAWFUL_DETENTION_1` and navigate to update-date-and-time — time field should be absent, hidden input `name="incidentTime" value="00:00"` should be present, and submitting should call the API with `00:00` as the time component

### DIFF
--- a/server/reportConfiguration/constants/types.ts
+++ b/server/reportConfiguration/constants/types.ts
@@ -359,7 +359,7 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
  * When set, the time field is hidden and this value is submitted to the API automatically.
  */
 export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
-  UNLAWFUL_DETENTION_1: '23:59',
+  UNLAWFUL_DETENTION_1: '00:00',
 }
 
 /**
--- a/server/reportConfiguration/constants/types.ts
+++ b/server/reportConfiguration/constants/types.ts
@@ -351,8 +351,15 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
   },
   UNLAWFUL_DETENTION_1: {
     incidentDate: 'The date the person was meant to be released - for example, 17/5/2024',
-    incidentTime: "For example, 09 08. If you're unsure, enter 23 59",
   },
+}
+
+/**
+ * Default times (HH:MM) for incident types where time input is not shown to the user.
+ * When set, the time field is hidden and this value is submitted to the API automatically.
+ */
+export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
+  UNLAWFUL_DETENTION_1: '23:59',
 }
 
 /**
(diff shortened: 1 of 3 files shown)


server/routes/reports/details/incidentDateAndTimeController.ts
```diff
@@ -22,12 +22,12 @@ export abstract class BaseIncidentDateAndTimeController<
...
   middlewareLocals(): void {
-    this.use(this.customiseIncidentDateHint)
+    this.use(this.customiseIncidentDateFields)
     super.middlewareLocals()
   }
 
-  /** Rewrite hint text for incident date if required */
-  private customiseIncidentDateHint(
+  /** Rewrite hint text and hide time field for incident date/time if configured for this type */
+  private customiseIncidentDateFields(
```
xoen: This is no longer just about the date field.

So we either rename it to something like `customiseIncidentDateAndTimeFields` or we split into separate function.

server/reportConfiguration/constants/types.ts
```diff
@@ -351,10 +351,17 @@ export const incidentTypeHints: { [K in Type]?: Record<string, string> } = {
...
-    incidentTime: "For example, 09 08. If you're unsure, enter 23 59",
   },
 }
 
+/**
+ * Default times (HH:MM) for incident types where time input is not shown to the user.
+ * When set, the time field is hidden and this value is submitted to the API automatically.
+ */
+export const incidentTypeDefaultTimes: Partial<Record<Type, string>> = {
+  UNLAWFUL_DETENTION_1: '00:00',
```
xoen: I can't shake the "feeling" this is not about having a default time but rather about the ability to not ask for the time, e.g. because you care about just the date.

This should be something like `incidentTypeRequiresTime: Partial<Record<Type, boolean>>` and you can override this to `false` when the time is not required (and that would default the time to `00:00`).

The default time is just a consequence of the fact the time is irrelevant for this incident type.

xoen (changes requested):
I need to have a proper look but I left some suggestions that may lead to tweaks to the rest of the PR.

server/reportConfiguration/constants/types.ts
Mjwillis: Yes agree, I originally did this so that the time should be set (e.g 23:59) but I think in reality it wouldn't be used

PR #857, opened 2026-05-27
IR-1646: Make TypeScript stricter and other smaller tidying ups

xoen:
Enabled more of [the `strict`-family](https://www.typescriptlang.org/tsconfig/#strict) TypeScript options:
- `strictBindCallApply`
- `strictBuiltinIteratorReturn`
- `noImplicitThis`
- `useUnknownInCatchVariables`

There are a few more options that are currently not enabled:
- [`alwaysStrict`](https://www.typescriptlang.org/tsconfig/#alwaysStrict), tests pass but it changes runtime behaviour so we may want to enabled separately
- [`strictNullChecks`](https://www.typescriptlang.org/tsconfig/#strictNullChecks), enable this currently throws 243 more errors in 48 files (some in tests).
- [`strictFunctionTypes`](https://www.typescriptlang.org/tsconfig/#strictFunctionTypes), enable this currently throws 6 errors in 4 files.
- [`strictPropertyInitialization`](https://www.typescriptlang.org/tsconfig/#strictPropertyInitialization), cannot be enabled without enabling `strictNullChecks`
- as a comparison, currently enabling `strict` in main throws 433 errors in 86 files (so these changes get rid of ~40% of the current type errors).

Ideally we'll want to get to a point where we can just enable `strict` and we don't need to enabled the other options individually.

However the changes to this PR reduced the number of type errors (and possibly logical errors).

Some highlights of the changes:
- removed unused/unnecessary nunjucks filters (there may be more)
- better checking of exceptions (in JS you can throw literally anything, but we were assuming the caught error had a certain shape)
- removed unnecessary type assertions (we should try and avoid them as much as possible as it's lying to TS and to ourselves)
- various changes to functions/method signatures to reflect the actual implementation (e.g. some functions were getting a parameter `T` but the body of the function was checking for things like `null | undefined` too)
- formalised dependency to certain `res.locals.*` variables
  - there are a number of middleware that when configured correctly can set things like `res.locals.report` or `res.locals.reportConfig` etc
  - the place where these are used was just assuming these were correctly configured, however TS was not convinced these couldn't be `undefined`
  - I've formalised this dependency and the check (so that TS is now narrowing the inferred type from `T | undefined` to `T`)
  - and when for whatever reason a variable in `res.local.*` was undefined we now get an explicit error telling us which middleware was not correctly configured (as this is a programming error)
  - this is a bit more verbose but at the same time, it's explicit and once you go past these checks you can have more confidence all your prerequisites are met
  - a similar check 
- better general handling of `null`/`undefined`
- applied some type-related changes from the TS template/[this PR](https://github.com/ministryofjustice/hmpps-template-typescript/pull/718) (there are possibly more that could be used to further reduce number of type errors)

Really, the high level, general principles should be:
- avoid type assertions as much as possible
- be liberal in what you accept, but conservative in what you produce/send
- validate and handle edge cases at the boundaries


scripts/drawStatusDiagrams.ts
```diff
@@ -32,12 +32,12 @@ function main(): void {
       }))
       .filter(({ actions }) => actions)
       .forEach(({ userType, userDescription, actions }) => {
-        const transitionEdges = Object.entries(actions)
+        const transitionEdges = Object.entries(actions!)
```
xoen, line 35: null assertions are not great generally, but here TS is confused and doesn't see through the `filter()` above (which filters out items with undefined `actions`).

There may be a better way of writing this but it will possibly require changing it to avoid using `forEach`/`filter` (which confuses TS) so maybe not ideal neither.

server/middleware/populateReport.ts
```diff
@@ -36,7 +36,7 @@ export function populateReport(withDetails: boolean): RequestHandler {
       next()
     } catch (error) {
       logger.error(error, `Failed to load report ${reportId}`)
-      if (error.responseStatus === 400) {
+      if (error && typeof error === 'object' && 'responseStatus' in error && error?.responseStatus === 400) {
```
jonesj2193, line 39: Very minor comment, but isn't `'responseStatus' in error` inherently covered in the following condition of `error?.responseStatus === 400`? As if it doesn't exist, it won't equal it. 
As I say, minor and not saying it needs to change as it's fine, just more of a comment that could reduce number of conditions/complexity in logic.

xoen: Kind of, but great observation!

From the point of view of the if condition I think you're right and the `'responseStatus' in error` should be redundant given the `error?.responseStatus === 400`.

However, the `'responseStatus' in error` is doing something a bit stronger, it's checking that the error has that property (the error could be **anything**, it starts as `unknown` and this is the type is narrowed down).
In this case it is mainly for TS benefit, otherwise it complains about the missing property (`Property 'responseStatus' does not exist on type 'object'.ts(2339)`). It is verbose but in JS that's the way you check for the shape of a value (bear in mind inside the block `error.responseStatus` is also updated).

In other situations the `'responseStatus' in error` would also save us from some logical errors, e.g. if we assume the property is there but in reality it isn't, for example:

```TypeScript
} catch (error) { // error is something like `{anything: 'goes'}`
  if (error && typeof error === 'object' && error?.responseStatus !== 200) {
     // do something
  }
```

without TS complaining we would have an error because we assumed error had the `responseStatus` property but it doesn't, so its value is `undefined` and `error?.responseStatus !== 200` returns `true` (in that case is probably best to re-throw/propagate that error because it's not the kind of error we were trying to handle).

Hope it makes sense!

jonesj2193 (approved):
LGTM

PR #859, opened 2026-05-27
IR-1654 Hide 'add another prisoner?' radio when all role slots are exhausted

Mjwillis:
## Problem

When a report type has prisoner roles where every active role has `onlyOneAllowed: true` and all those roles are already filled, the prisoner involvement summary page continued to show the **"Do you want to add another prisoner? Yes / No"** radio button.

Choosing **Yes** sent the user through the full prisoner search flow, only to be rejected at the end with:

> **Error — No more prisoner roles can be added**
> You may need to remove an existing person.

The service had all the information it needed to know no further prisoners could be added; asking the question was pointless and created a frustrating dead-end.

Affected incident types include any type where every active prisoner role has `onlyOneAllowed: true`, for example:
- `ABSCOND_1` — sole role: `ABSCONDER`
- `FOOD_REFUSAL_1` — sole role: `PERPETRATOR`
- `DIRTY_PROTEST_1`, `UNLAWFUL_DETENTION_1`
- Any type once every available role slot has been filled (e.g. `ESCAPE_FROM_PRISON_1` once the single `ESCAPE` role is taken, even though other `onlyOneAllowed: false` roles remain)

## Solution

Detect the exhausted-roles state on the summary page and suppress the radio, so the **Continue** and **Save and exit** buttons act as an implicit "No". If the existing prisoner is removed, the available role slot opens back up and the radio reappears on the next visit.

The check mirrors the logic already present in `PrisonerInvolvementController.getAllowedPrisonerRoles()` (used to produce the error): a role slot is still available if `!onlyOneAllowed` (unlimited) or `onlyOneAllowed && !alreadyUsed`.

### `server/routes/reports/prisoners/summary/controller.ts`
- Added `middlewareLocals()` override that prepends `populateReportConfiguration(false)` via `this.router.use()` before calling `super.middlewareLocals()`. This mirrors the identical pattern in `PrisonerInvolvementController` and makes `res.locals.reportConfig` available on the summary page for the first time.
- Added `protected canAddMoreInvolvements(res)`: iterates `reportConfig.prisonerRoles` against the involvements already on the report to return `false` when every active role is `onlyOneAllowed: true` and already used.

### `server/controllers/involvements/summary.ts` (abstract base)
- Added `protected canAddMoreInvolvements(_res): boolean { return true }` — the safe default ensures the staff summary page (which has no `onlyOneAllowed` concept) is completely unaffected.
- In `customiseFields()`: calls `canAddMoreInvolvements()`. When `involvementsExist && !canAddMore`, the `validate` array on the `confirmAdd` field is cleared (removing the `required` rule so form-wizard does not reject an empty POST) and `res.locals.noMoreCanBeAdded` is set to `true`.
- No change to `successHandler()`: an absent/undefined `confirmAdd` already falls through the `else` branch (not `'yes'`), and the `prisonerInvolvementDone` API update is only triggered when `confirmAdd === 'no' && involvements.length === 0` — neither condition applies here.

### `server/views/pages/involvements/summary.njk`
- Wrapped the entire `govukRadios(…)` call in `{% if not noMoreCanBeAdded %}…{% endif %}`.
- The Continue and Save and exit buttons remain **unconditional**.

### Unit tests — `server/routes/reports/prisoners/summary/index.test.ts`
New `describe` block using a report of type `ABSCOND_1` (the simplest case: one role, `ABSCONDER`, `onlyOneAllowed: true`) with that role already filled:

| Test | What it covers |
|---|---|
| GET — radio absent, buttons present (create + normal) | visual suppression |
| POST empty body — correct redirect (create / exit / normal) | implicit No navigation |
| POST empty body — `updateReport` not called | no spurious API write (prisoners exist) |
| POST empty body — 302, not validation error | `validate: []` clears required rule |

All 37 pre-existing tests continue to pass unchanged; `FIND_6` (the default test type) has only `onlyOneAllowed: false` roles so the radio still shows as before.

### Cypress integration tests — `integration_tests/e2e/involvements/prisoners/involvements.cy.ts`
New `context` block with `ABSCOND_1` + one `ABSCONDER` prisoner:
- Radio is absent (`showsNoRadio()` helper added to `InvolvementsPage`)
- Continue and Save and exit buttons exist
- Clicking Continue navigates to the report page

### DIFF
--- a/server/controllers/involvements/summary.ts
+++ b/server/controllers/involvements/summary.ts
@@ -35,6 +35,14 @@ export abstract class InvolvementSummary extends BaseController<Values> {
     super.middlewareLocals()
   }
 
+  // Override in subclasses to prevent the "add another?" radio from showing when
+  // no further involvements can be added (e.g. all prisoner roles with onlyOneAllowed
+  // are already used).  Default returns true so staff summary is unaffected.
+
+  protected canAddMoreInvolvements(_res: express.Response): boolean {
+    return true
+  }
+
   private customiseFields(req: FormWizard.Request<Values>, res: express.Response, next: express.NextFunction): void {
     const report = res.locals.report as ReportWithDetails
 
@@ -53,10 +61,19 @@ export abstract class InvolvementSummary extends BaseController<Values> {
     if (involvementsExist) {
       customisedFields.confirmAdd.label = this.labelOnceInvolvementsExist
     }
+
+    const canAddMore = this.canAddMoreInvolvements(res)
+    if (involvementsExist && !canAddMore) {
+      // Remove required validation: the radio will be hidden in the template and
+      // an empty submission is treated as an implicit "No" in successHandler.
+      customisedFields.confirmAdd = { ...customisedFields.confirmAdd, validate: [] }
+    }
+
     req.form.options.fields = customisedFields
 
     res.locals.involvementDone = involvementDone
     res.locals.involvementsExist = involvementsExist
+    res.locals.noMoreCanBeAdded = involvementsExist && !canAddMore
 
     next()
   }
--- a/server/controllers/involvements/summary.ts
+++ b/server/controllers/involvements/summary.ts
@@ -63,7 +63,8 @@ export abstract class InvolvementSummary extends BaseController<Values> {
     }
 
     const canAddMore = this.canAddMoreInvolvements(res)
-    if (involvementsExist && !canAddMore) {
+    const noMoreCanBeAdded = involvementsExist && !canAddMore
+    if (noMoreCanBeAdded) {
       // Remove required validation: the radio will be hidden in the template and
       // an empty submission is treated as an implicit "No" in successHandler.
       customisedFields.confirmAdd = { ...customisedFields.confirmAdd, validate: [] }
@@ -73,7 +74,7 @@ export abstract class InvolvementSummary extends BaseController<Values> {
 
     res.locals.involvementDone = involvementDone
     res.locals.involvementsExist = involvementsExist
-    res.locals.noMoreCanBeAdded = involvementsExist && !canAddMore
+    res.locals.noMoreCanBeAdded = noMoreCanBeAdded
 
     next()
   }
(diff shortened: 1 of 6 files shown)


jonesj2193 (dismissed):
LGTM

server/controllers/involvements/summary.ts
```diff
@@ -53,10 +61,19 @@ export abstract class InvolvementSummary extends BaseController<Values> {
...
+      // Remove required validation: the radio will be hidden in the template and
+      // an empty submission is treated as an implicit "No" in successHandler.
+      customisedFields.confirmAdd = { ...customisedFields.confirmAdd, validate: [] }
+    }
+
     req.form.options.fields = customisedFields
 
     res.locals.involvementDone = involvementDone
     res.locals.involvementsExist = involvementsExist
+    res.locals.noMoreCanBeAdded = involvementsExist && !canAddMore
```
xoen: you could probably move this in a constant before the `if` as this is the same as the `if` condition.

server/routes/reports/prisoners/summary/controller.ts
```diff
@@ -2,13 +2,38 @@ import type express from 'express'
...
 export default class PrisonerSummary extends InvolvementSummary {
+  middlewareLocals(): void {
+    // Load reportConfig (without generating question steps) so canAddMoreInvolvements()
+    // can inspect prisonerRoles.  Mirrors the pattern in PrisonerInvolvementController.
+    this.router.use(populateReportConfiguration(false))
+    super.middlewareLocals()
+  }
+
+  protected canAddMoreInvolvements(res: express.Response): boolean {
+    const reportConfig = res.locals.reportConfig as IncidentTypeConfiguration | undefined
```
xoen: I don't think you need this type assertion here, `res.locals.reportConfig` is already `IncidentTypeConfiguration | undefined` (and you have the if below that should narrow it down to `IncidentTypeConfiguration`)

PS: In general we should try to avoid type assertions (`as T`) as much as possible as they're a way to lie to TS and can mask possible bugs.

Mjwillis: @xoen pushed changes

PR #863, opened 2026-06-01
IR-1657 Auto-skip the prisoner-role radio when only one role can be chosen

Mjwillis:
# IR-1657: Auto-assign the prisoner's role when only one is possible

## What & why

When staff add or edit a prisoner on an incident report, they pick the prisoner's **role** from a radio list. For some incident types the role list only ever offers a single choice — e.g. `UNLAWFUL_DETENTION_1` defines exactly one active role (`UNLAWFULLY_DETAINED`). Forcing a click on the sole radio button is a pointless step.

This change removes that click: when only one selectable role remains, the role is assigned automatically and submitted via a hidden field. For clarity, the assigned role is shown **read-only** above the details box so staff can see exactly which role will be recorded — they just don't have to select it.

The page then asks only for the optional involvement details (plus the outcome field for NOMIS-origin reports, unchanged).

This mirrors the existing IR-1651 optimisation, which does the same auto-skip for single-answer *questions*.

## Behaviour

- **Trigger:** whenever the *runtime* list of selectable roles reduces to exactly one. This covers single-role incident types and multi-role types whose `onlyOneAllowed` slots have been exhausted down to one.
- **Add:** no role radio; a read-only "Role" line shows the assigned role; only the optional details box is editable. The role is persisted via the hidden field.
- **Edit:** the role stays hidden/read-only and the existing details pre-fill, so reporters can amend details without re-selecting the role.
- **Multi-role types:** unchanged — the radio still appears whenever more than one role is selectable.

## Changes

| File | Change |
|---|---|
| `server/routes/reports/prisoners/involvement/controller.ts` | In the shared `customisePrisonerRoles()`, when exactly one role remains, switch the `prisonerRole` field to `component: 'hidden'` with `default` set to that role. Covers both add and edit (shared base class). |
| `server/views/macros/renderField.njk` | The generic `hidden` branch now falls back to `field.default` (mirrors `questions.njk`). Backwards-compatible — other hidden fields set no default. |
| `server/views/pages/prisoners/involvement.njk` | Override the `formFields` block to render a read-only GOV.UK summary-list row (`Role` → role description) when the role is auto-assigned, then render the normal fields (which still emit the hidden input + outcome/comment). |

## Tests

New single-role blocks in `add.test.ts` and `edit.test.ts` (using `UNLAWFUL_DETENTION_1`):
- GET hides the role radio, shows the role read-only, and emits the hidden input.
- POST persists the auto-assigned role with the submitted details.
- Edit keeps the role hidden, pre-fills existing details, and saves updated details with the role unchanged.

Existing multi-role tests are untouched, proving the radio still shows when there's a genuine choice.

## Verification

- `npm test` — all involvement suites pass (116)
- `npm run typecheck` — clean
- `npm run lint` — clean

## Manual testing

Create an `UNLAWFUL_DETENTION_1` report → add a prisoner → confirm no role radio, the role shown read-only, and only the details box editable. Save, then Edit the prisoner → role stays read-only, details pre-fill. Compare with a multi-role type (e.g. a FIND report) → radio still appears.


jonesj2193:
Having looked to double check what the page looks like now, the page does not now show which role is being allocated to the prisoner in question. Whilst we can remove the need to select a role, we may want to still show the user the role being selected. As if there are multiple roles to allocate to prisoners for a report, which one that is left becomes dependent on the order you have added them, so it will become unclear which role you are now allocated to the person you are currently adding. 
Hope that makes sense? happy to discuss further.

jonesj2193 (approved):
I think the change to display it is better and clearer to the users. May want to change the display to just be text of "Role: defined_role" or something else in future. But happy to merge this in as is and tweak again in future if needed.

PR #866, opened 2026-06-02
IR-1643 checkbox hint text

jonesj2193:
It was raised that is may be unclear to users the difference between radio button and checkbox questions, and they may not be aware they can select multiple options when checkboxes appear. 

This PR adds in to the question template that if the question is a "multiple answers" question and does not have any predefined hint text associated with it, the hint test will be populated with "Select all that apply". 

Due to this change, "(select all that apply)" was removed from the labels for two questions and the new hint text renders it redundant.


Mjwillis (dismissed):
LGTM

server/data/testData/FIND_6.ts
```diff
@@ -15,7 +15,7 @@ export const config: IncidentTypeConfiguration = {
       code: '67179',
       active: true,
       question: 'DESCRIBE HOW THE ITEM WAS FOUND (SELECT ALL THAT APPLY)',
-      label: 'Describe how the item was found (select all that apply)',
+      label: 'Describe how the item was found',
```
xoen, line 18: would you mind re-generating the graphs?

server/views/macros/renderCheckboxOrRadioField.njk
```diff
@@ -42,6 +42,12 @@
 
   {% if field.multiple %}
 
+    {% if not field.hint %}
```
xoen: you could flip the condition and avoiding the `not`.

server/routes/reports/questions/index.test.ts
```diff
@@ -95,6 +99,32 @@ describe('Displaying questions and responses', () => {
         })
     })
 
+    it('multiple choices questions display "Select all that apply" hint text if no hint text provided', () => {
```
xoen, line 98: Usually the set up of the test is done in a `beforeEach|All` block (when it makes sense), for example:

```TypeScript
describe('multiple choices questions, if no hint text provided', () => {
  beforeEach(() => {
    reportWithDetails.type = 'FIND_6'
  })

  it('display "Select all that apply" as hint text', () => {
    // ...
  })
})

describe('multiple choices questions, when hint text is defined', () => {
  beforeEach(() => {
    reportWithDetails.type = 'FIND_6'
    FIND_6.questions['67179'].questionHint = 'Bespoke hint text'
  })

  it('display bespoke hint text instead of "Select all that apply"', () => {
    // ...
  })
})
```

and the output of the test run can be like live documentation, e.g.

```
[...]
  multiple choices questions, if no hint text provided
    ✓ display "Select all that apply" as hint text (83 ms)
  multiple choices questions, when hint text is defined
    ✓ display bespoke hint text instead of "Select all that apply" (83 ms)
```

jonesj2193: But in this instance that part of the set up is only applicable to this 1 specific unit test. So I would have to take this test out of the current stack of tests and have it by itself. Which then would actually result in more duplicated code. Hence why I did it this way.

server/data/testData/FIND_6.ts
jonesj2193: done this now

server/views/macros/renderCheckboxOrRadioField.njk
jonesj2193: But the exception/rule we are looking for is the absence of hint text. So this is the simplest version of the logic to apply it when needed. Trying to flip it to simply avoid using not would actually make it more convoluted/confusing. I think it's very clear what is going on as it is.

xoen: I don't think it's too bad as it is, but I still think `if not X` is less clear than `if X`, the two branches could be easily swapped around and it would not make it any more confused (and definitely not convoluted).

xoen (dismissed):
Looks good, left some minor comments.

xoen (approved):

PR #868, opened 2026-06-07
IR-1664 Date-based incident type activation + Food refusal v2 / Close down search retirement

Mjwillis:
# Date-based incident type activation + Food refusal v2 / Close down search retirement

## Why

Two business changes need to go live on **1 July 2026**:

1. **Food refusal** – the Safety team has reduced/changed the question set. Editing the existing
   `FOOD_REFUSAL_1` config in place would retro-actively alter or remove answers on reports that
   were already captured under it, potentially making them un-editable. So we need a **new version**
   (`FOOD_REFUSAL_2`) and a way to switch the family from v1 → v2 automatically on a date.
2. **Close down search** – decommissioned (no replacement); must stop being offered for new reports.

Until now "active" was a single boolean per incident type, which cannot express "active until / from
a date". This PR introduces **date-based activation windows** so that, per family, exactly one
version is active at any moment, with the switch-over happening automatically on a calendar date.

### Date-based activation
- New hand-maintained module `server/reportConfiguration/constants/typeActivePeriods.ts`:
    - `typeActivePeriods` – per-type `{ activeFrom?, activeTo? }` windows (ISO `YYYY-MM-DD`,
      Europe/London, half-open: `from` inclusive, `to` exclusive, so a retiring version's `activeTo`
      can equal its successor's `activeFrom` with no overlap/gap).
    - `isTypeActive(code, at?)` – a type is active **iff its generated `active` boolean is `true` AND
      `at` is within its window**. Dates can only *retire* or *delay* a type; they never reactivate a
      boolean-off type (important – several retired NOMIS types carry recent/future recorded end dates).
      Comparison is London date-only via `format.isoDate`, so switches land at local midnight (BST/GMT
      safe).
- Wired into the three runtime "is this type usable" sites, replacing the raw boolean:
  `routes/reports/details/typeFields.ts` (new-report list, also feeds change-type),
  `middleware/populateReportConfiguration.ts`, and `routes/reports/questions/index.ts`.
- Windows added: `FOOD_REFUSAL_1` & `CLOSE_DOWN_SEARCH_1` → `activeTo: 2026-07-01`;
  `FOOD_REFUSAL_2` → `activeFrom: 2026-07-01`. Plus recorded historical end dates for all 25
  already-retired types (metadata only – boolean already false). No type needed a fallback date.

### Food refusal v2
- New type `FOOD_REFUSAL_2` added to the registry (`constants/types.ts` + `typesDescriptions`),
  `nomisCode: 'FOOD_REF2'` (config is back-synced to NOMIS separately via
  `scripts/buildNomisIncidentTypePayload.ts`).
- `types/FOOD_REFUSAL_2.ts` (+ regenerated `.dot`/`.svg`) is a clone of `FOOD_REFUSAL_1`'s full
  question flow (all 10 questions / 31 answers) with **only the two changed questions**:
    - *"Which type of refusal occurred?"* → Food / Liquid / Food and liquid (Other removed).
    - *"Where is the prisoner currently located?"* → adds **Constant supervision cell**; keeps Normal
      location, Segregation unit, Health care centre, Outside hospital, Other (free text).
    - PERPETRATOR only (one allowed); at least one prisoner required (as v1).
- **Every question and answer code is freshly allocated and globally unique** (questions `900001+`,
  answers `950001+`). This is required because the codes are NOMIS primary keys – reusing v1 codes
  would cause PK clashes on back-sync.

### Dev preview override
- New env var `INCIDENT_TYPE_ACTIVE_DATE` (`config.incidentTypeActiveDate`). When set, `isTypeActive`
  uses it as the effective "today" (empty = real now).
- Set to `"2026-07-01"` in `helm_deploy/values-dev.yaml`, so **dev previews the go-live state now**:
  Food refusal resolves to v2, and v1 + Close down search are hidden – without waiting for 1 July.
  The existing `INCIDENT_TYPES_OVERRIDE` force-include mechanism is unchanged.

## Tests
- `constants/typeActivePeriods.test.ts` – switch-over behaviour, boundary semantics (`from`
  inclusive / `to` exclusive), "never active when the boolean is false", the
  **one-active-version-per-family invariant** (sampled at `now` and every window boundary ±1 day),
  and the `INCIDENT_TYPE_ACTIVE_DATE` override.
- `data/incidentTypeConfiguration/validation.test.ts` – new **global code-uniqueness** checks (no
  question/answer code shared between any two configs); the existing loop also validates
  `FOOD_REFUSAL_2`.
- `routes/reports/details/changeType.test.ts` – updated to mirror `isTypeActive`.

## Verification
- `npm run typecheck` ✅  `npm run lint` ✅
- Full unit suite **2668 passed** (run with `--runInBand`; the parallel runner has a pre-existing
  flaky failure in `prisoners/involvement/add.test.ts` unrelated to this change – passes in isolation
  and serially).

## Notes / follow-ups
- The two changed questions use new machine text/responses (`WHICH TYPE OF REFUSAL OCCURRED`;
  `FOOD` / `LIQUID` / `FOOD AND LIQUID`; `CONSTANT SUPERVISION CELL`). Confirm these match what NOMIS
  expects before back-sync.
- The dev override shifts *all* date logic in dev (intended – it previews the whole 1 July state).


server/reportConfiguration/constants/typeActivePeriods.ts
```diff
@@ -0,0 +1,102 @@
...
+ * A type is only ever active when its registry boolean is `true` AND now falls within its
+ * window (see {@link isTypeActive}). Windows therefore can only *retire* or *delay* a type;
+ * a recorded `activeTo` on an already-inactive (boolean `false`) type is metadata only and
+ * never reactivates it — important because some retired NOMIS types carry recent/future end
+ * dates that are not monotonic by version number.
+ */
+export const typeActivePeriods: Partial<Record<Type, ActivePeriod>> = {
+  // Family switch-over for 1 July 2026: Food refusal v1 retires as v2 begins.
+  FOOD_REFUSAL_1: { activeTo: '2026-07-01' },
+  FOOD_REFUSAL_2: { activeFrom: '2026-07-01' },
```
xoen: It would be good to have these keys sorted alphabetically, mainly for our benefit when looking at/updating this.

server/reportConfiguration/types/FOOD_REFUSAL_2.ts
```diff
@@ -0,0 +1,132 @@
...
+    '900001': {
+      code: '900001',
+      active: true,
+      question: 'WHICH TYPE OF REFUSAL OCCURRED',
+      label: 'Which type of refusal occurred?',
+      multipleAnswers: false,
+      answers: [
+        {
+          code: '950001',
+          response: 'FOOD',
```
xoen: We could now change the actual responses to _not_ be SHOUT-case no?

The new version breaks with the past anyway, what's the rational for keeping the original responses? 🤔

server/reportConfiguration/types/index.ts
```diff
@@ -8,10 +8,17 @@ export function getAllIncidentTypeConfigurations(): IncidentTypeConfiguration[]
...
 
-  return import(configPath).then(module => module.default.default)
+  // Depending on the runtime (ts-jest, tsx, built JS), CJS/ESM interop nests the
+  // default export at a different depth — unwrap `.default` until we reach the config.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let config: any = await import(configPath)
+  while (config && typeof config === 'object' && !('incidentType' in config) && 'default' in config) {
+    config = config.default
+  }
+  return config as IncidentTypeConfiguration
```
xoen: What's going on here? 🤔

xoen (changes requested):
This looks great, I had a few questions/potential tweaks (e.g. responses).
But mainly I don't quite get the change to the loading of the incident report config, is that needed?

server/data/incidentTypeConfiguration/validation.test.ts
```diff
@@ -17,6 +17,46 @@ describe('Active incident type configurations', () => {
   })
 })
 
+describe('Question and answer codes are globally unique', () => {
+  // NOMIS stores question/answer codes as primary keys, so a code must never be shared between two
+  // type configs (e.g. a new version must use brand-new codes, not reuse the previous version's).
+  const allConfigs = getAllIncidentTypeConfigurations()
+
+  it('has no question code used by more than one incident type', () => {
+    const owners = new Map<string, string>()
```
xoen, line 26: owners?

xoen: oh as in "the config that owns the question/answer code" presumably

server/reportConfiguration/types/index.ts
Mjwillis: `./scripts/buildNomisIncidentTypePayload.ts FOOD_REFUSAL_2` crashed with
`TypeError: Cannot read properties of undefined (reading 'questions')` because
`getIncidentTypeConfiguration()` resolved to `undefined`.

The loader in `server/reportConfiguration/types/index.ts` dynamically imported the
type config and reached into it with a hard-coded double default:

```ts
return import(configPath).then(module => module.default.default)
```

That exact `.default.default` depth only lands on the config under some runtimes.
The script runs via **tsx**, which injects its own `__esModule` CJS/ESM interop that
already unwraps one `.default` level — so `.default.default` went one level too deep
and returned `undefined`. Under ts-jest / built JS the same code happened to work,
which is why the regression was invisible until the script was run.

## What changed

- `getIncidentTypeConfiguration()` now unwraps `.default` iteratively until it reaches
  the actual config object (identified by its `incidentType` field), instead of
  assuming a fixed nesting depth. This is robust across tsx, ts-jest and built JS:

  ```ts
  let config: any = await import(configPath)
  while (config && typeof config === 'object' && !('incidentType' in config) && 'default' in config) {
    config = config.default
  }
  return config as IncidentTypeConfiguration
  ```

This loader is also used at runtime by `populateReportConfiguration` middleware, so the
fix applies everywhere the function is called, not just the script.

Mjwillis:
> This looks great, I had a few questions/potential tweaks (e.g. responses). But mainly I don't quite get the change to the loading of the incident report config, is that needed?

See comments

server/reportConfiguration/types/FOOD_REFUSAL_2.ts
Mjwillis: No real reason, I can update

xoen (approved):
Again, really nice work ❤️

PR #869, opened 2026-06-09
IR-1673 Prisoner incident summary page

Mjwillis:
## What & why

Adds a new read-only page that, for a given prisoner, summarises the incidents they have been
involved in over the past 12 months. It is intended to be linked to from other HMPPS services,
so unlike the rest of this app it is **accessible to any signed-in user**, provided the prisoner
is in one of their caseloads.

The page shows:

- **All incidents by type** — a count of (non-excluded) reports per incident-type family.
- **Detail subsections** for Assault, Disorder, Self-harm and Find, drilling into each report's
  questions/responses for extra sub-counts.

URL: `GET /prisoner/:prisonerNumber/incident-summary`

prisoner-incident-summary

## Behaviour & decisions

- **12-month window** is measured against the **incident date** (`incidentDateFrom = today − 12
  months`).
- **Excluded statuses**: `DRAFT`, `DUPLICATE`, `NOT_REPORTABLE`. All other statuses are included
  (derived from the `statuses` constant so it stays correct if new statuses are added).
- **Caseload access**: a prisoner outside the user's caseloads returns **404 Not Found**, not
    403. The app's error handler turns 403 into a sign-out, which is wrong for a legitimately
         signed-in user; 404 also avoids confirming a prisoner exists. This naturally denies out/transfer
         prisoners (prisonId `OUT`/`TRN`).
- **Detail extraction covers active type versions only** — `ASSAULT_5`, `DISORDER_2`,
  `SELF_HARM_1`, `FIND_6`. Reports of older versions are still counted in the overall/family
  totals but are not drilled into. A short doc comment in `breakdowns.ts` explains how to add a
  new version when one becomes active.
- A report can fall into several sub-buckets (questions allow multiple answers), so sub-counts are
  report-counts and are not expected to sum to the family total.

### Detail sub-counts

| Family | Sub-counts |
|---|---|
| Assault | Prisoner-on-prisoner / staff / other, serious injury, sexual, spitting, concussion/internal-injury treatment |
| Disorder | Barricade/prevention of access, concerted indiscipline, hostage, incident at height |
| Self-harm | Cutting, strangulation **or** hanging (combined), burning, other method |
| Find | Categories: alcohol / drugs / weapons; drug sub-types: cannabis, cocaine, NPS, other, unknown |

Find category detection covers both the single-select "category of find" path and the "multiple
types" path (which records each category as its own question). Drug sub-types are matched on the
response text because the codes differ between the two "describe the drug found" questions.
Drug grouping assumption: **CRACK → cocaine**, **CANNABIS PLANT → cannabis**.

## Implementation

The summary logic is a plain function module (not a registered singleton service) because the
`incidentReportingApi` client is built per-request in `setApis`; the controller passes
`res.locals.apis.incidentReportingApi` into it.

| File | Purpose |
|---|---|
| `server/services/prisonerIncidentSummary/breakdowns.ts` | Config-driven per-active-type extractors + helpers; response codes commented against their source config file |
| `server/services/prisonerIncidentSummary/index.ts` | `getPrisonerIncidentSummary(api, prisonerNumber)` — pages through `getReports`, groups by family, fetches `with-details` only for active-version reports in the four special families |
| `server/middleware/requirePrisonerInCaseload.ts` | Guards the route (404 when prisoner not in caseload); runs after `populatePrisoner()` |
| `server/routes/prisonerIncidentSummary/index.ts` | Router: `populatePrisoner()` → `requirePrisonerInCaseload()` → controller |
| `server/views/pages/prisonerIncidentSummary.njk` | GOV.UK page: overall table, per-family detail tables, empty state |
| `server/routes/index.ts` | Mounts the router at `/prisoner/:prisonerNumber/incident-summary` |
| `server/@types/express/index.d.ts` | Adds `prisoner?: OffenderSearchResult` to `res.locals` (set by `populatePrisoner`) |

Reuses existing building blocks: `IncidentReportingApi.getReports` / `getReportWithDetailsById`,
`getTypeDetails` / `typeFamiliesDescriptions`, the `statuses` constant, `populatePrisoner()`,
and the caseload pattern from `permissions/rulesClass.ts`.

## Tests

- `breakdowns.test.ts` — extractor logic, incl. multi-answer, non-YES, multiple-types find path,
  NIL-quantity alcohol, and drug grouping across both drug questions.
- `index.test.ts` (service) — status filter, ~12-month boundary, family grouping/ordering,
  pagination, older-version reports counted but not drilled into, find breakdown.
- `requirePrisonerInCaseload.test.ts` — allow / 404 (out of caseload) / 404 (out-of-prison) / 500
  (middleware misordered).
- `routes/prisonerIncidentSummary/index.test.ts` — renders for in-caseload prisoner, empty state,
  404 for out-of-caseload.
- `integration_tests/e2e/prisonerIncidentSummary.cy.ts` — Cypress test stubbing all four special
  families plus plain ones, asserting the rendered content and capturing a screenshot.

### DIFF
--- a/integration_tests/e2e/prisonerIncidentSummary.cy.ts
+++ b/integration_tests/e2e/prisonerIncidentSummary.cy.ts
@@ -1,0 +1,67 @@
+import type { ReportWithDetails } from '../../server/data/incidentReportingApi'
+import { mockReport } from '../../server/data/testData/incidentReporting'
+import { makeSimpleQuestion } from '../../server/data/testData/incidentReportingJest'
+import { andrew } from '../../server/data/testData/offenderSearch'
+import { mockReportingOfficer } from '../../server/data/testData/users'
+import { now } from '../../server/testutils/fakeClock'
+
+/** Build a closed report of the given type with crafted questions, all within the last 12 months. */
+function reportOfType(reference: string, type: string, questions: ReportWithDetails['questions']): ReportWithDetails {
+  const report = mockReport({
+    reportReference: reference,
+    reportDateAndTime: now,
+    type: type as ReportWithDetails['type'],
+    status: 'CLOSED',
+    location: andrew.prisonId,
+    withDetails: true,
+  }) as unknown as ReportWithDetails
+  report.questions = questions
+  return report
+}
+
+describe('Prisoner incident summary', () => {
+  it('shows the overall and per-family breakdowns', () => {
+    const assaultA = reportOfType('6001', 'ASSAULT_5', [
+      makeSimpleQuestion('61287', 'WHAT TYPE OF ASSAULT WAS IT', ['PRISONER ON PRISONER', '213115']),
+      makeSimpleQuestion('61285', 'WAS THIS A SEXUAL ASSAULT', ['YES', '1']),
+      makeSimpleQuestion('61290', 'WAS SPITTING USED IN THIS INCIDENT', ['YES', '2']),
+    ])
+    const assaultB = reportOfType('6002', 'ASSAULT_5', [
+      makeSimpleQuestion('61287', 'WHAT TYPE OF ASSAULT WAS IT', ['PRISONER ON STAFF', '213116']),
+      makeSimpleQuestion('61298', 'WAS A SERIOUS INJURY SUSTAINED', ['YES', '3']),
+      makeSimpleQuestion('61305', 'WAS MEDICAL TREATMENT FOR CONCUSSION OR INTERNAL INJURIES REQUIRED', ['YES', '4']),
+    ])
+    const disorder = reportOfType('6003', 'DISORDER_2', [
+      makeSimpleQuestion('63179', 'WHAT TYPE OF DISORDER INCIDENT WAS THIS?', ['HOSTAGE', '214686']),
+    ])
+    const selfHarm = reportOfType('6004', 'SELF_HARM_1', [
+      makeSimpleQuestion('44753', 'DID SELF HARM METHOD INVOLVE CUTTING', ['YES', '5']),
+      makeSimpleQuestion('45167', 'DID SELF HARM METHOD INVOLVE BURNING', ['YES', '6']),
+    ])
+    const find = reportOfType('6005', 'FIND_6', [
+      makeSimpleQuestion('67187', 'PLEASE SELECT CATEGORY OF FIND', ['DRUG / DRUG EQUIPMENT', '218785']),
+      makeSimpleQuestion('67190', 'DESCRIBE THE DRUG FOUND', ['CANNABIS', '7'], ['CRACK', '8']),
+    ])
+    const fire = reportOfType('6006', 'FIRE_1', [])
+    const misc = reportOfType('6007', 'MISCELLANEOUS_1', [])
+
+    const allReports = [assaultA, assaultB, disorder, selfHarm, find, fire, misc]
+    const detailedReports = [assaultA, assaultB, disorder, selfHarm, find]
+
+    cy.resetBasicStubs({ user: mockReportingOfficer })
+    cy.signIn()
+    cy.task('stubOffenderSearchMockPrisoners')
+    cy.task('stubIncidentReportingApiGetReports', { reports: allReports })
+    detailedReports.forEach(report => cy.task('stubIncidentReportingApiGetReportWithDetailsById', { report }))
+
+    cy.visit(`/prisoner/${andrew.prisonerNumber}/incident-summary`)
+
+    cy.get('h1').should('contain.text', 'Incident summary')
+    cy.contains('Assault').should('exist')
+    cy.contains('Prisoner on prisoner').should('exist')
+    cy.contains('Hostage').should('exist')
+    cy.contains('Cannabis').should('exist')
+
+    cy.screenshot('prisoner-incident-summary', { capture: 'fullPage' })
+  })
+})
--- a/integration_tests/e2e/prisonerIncidentSummary.cy.ts
+++ b/integration_tests/e2e/prisonerIncidentSummary.cy.ts
@@ -1,20 +1,24 @@
 import type { ReportWithDetails } from '../../server/data/incidentReportingApi'
+import { convertReportWithDetailsDates } from '../../server/data/incidentReportingApiUtils'
 import { mockReport } from '../../server/data/testData/incidentReporting'
 import { makeSimpleQuestion } from '../../server/data/testData/incidentReportingJest'
 import { andrew } from '../../server/data/testData/offenderSearch'
 import { mockReportingOfficer } from '../../server/data/testData/users'
+import type { Type } from '../../server/reportConfiguration/constants'
 import { now } from '../../server/testutils/fakeClock'
 
 /** Build a closed report of the given type with crafted questions, all within the last 12 months. */
-function reportOfType(reference: string, type: string, questions: ReportWithDetails['questions']): ReportWithDetails {
-  const report = mockReport({
-    reportReference: reference,
-    reportDateAndTime: now,
-    type: type as ReportWithDetails['type'],
-    status: 'CLOSED',
-    location: andrew.prisonId,
-    withDetails: true,
-  }) as unknown as ReportWithDetails
+function reportOfType(reference: string, type: Type, questions: ReportWithDetails['questions']): ReportWithDetails {
+  const report = convertReportWithDetailsDates(
+    mockReport({
+      reportReference: reference,
+      reportDateAndTime: now,
+      type,
+      status: 'CLOSED',
+      location: andrew.prisonId,
+      withDetails: true,
+    }),
+  )
   report.questions = questions
   return report
 }
--- a/server/middleware/requirePrisonerInCaseload.ts
+++ b/server/middleware/requirePrisonerInCaseload.ts
@@ -1,7 +1,8 @@
 import type { RequestHandler } from 'express'
-import { InternalServerError, NotFound } from 'http-errors'
+import { NotFound } from 'http-errors'
 
 import logger from '../../logger'
+import { missingLocalsError } from '../errors'
 
 /**
  * Guards a route so that it can only be accessed when the loaded prisoner is held in one of the
@@ -18,7 +19,7 @@ export function requirePrisonerInCaseload(): RequestHandler {
   return (_req, res, next): void => {
     const { prisoner, user } = res.locals
     if (!prisoner) {
-      next(new InternalServerError('requirePrisonerInCaseload() requires populatePrisoner() to run first'))
+      next(missingLocalsError('requirePrisonerInCaseload()', 'res.locals.prisoner'))
       return
     }
 
--- a/server/middleware/requirePrisonerInCaseload.ts
+++ b/server/middleware/requirePrisonerInCaseload.ts
@@ -1,0 +1,34 @@
+import type { RequestHandler } from 'express'
+import { InternalServerError, NotFound } from 'http-errors'
+
+import logger from '../../logger'
+
+/**
+ * Guards a route so that it can only be accessed when the loaded prisoner is held in one of the
+ * current user's caseloads (caseloadId === prisonId). Must run after populatePrisoner().
+ *
+ * This naturally denies prisoners who are out or in transit, as their prisonId ("OUT" / "TRN")
+ * will not match any caseload.
+ *
+ * A denied prisoner is treated as Not Found rather than Forbidden: the user is legitimately
+ * signed in (so the app's 403-triggers-sign-out behaviour would be wrong), and we avoid
+ * confirming that a prisoner outside their caseload exists.
+ */
+export function requirePrisonerInCaseload(): RequestHandler {
+  return (_req, res, next): void => {
+    const { prisoner, user } = res.locals
+    if (!prisoner) {
+      next(new InternalServerError('requirePrisonerInCaseload() requires populatePrisoner() to run first'))
+      return
+    }
+
+    const caseloadIds = new Set(user?.caseLoads?.map(caseLoad => caseLoad.caseLoadId) ?? [])
+    if (!caseloadIds.has(prisoner.prisonId)) {
+      logger.warn(`User denied access to prisoner ${prisoner.prisonerNumber}: not in caseload`)
+      next(new NotFound('Prisoner is not in your caseload'))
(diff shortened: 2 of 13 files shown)


xoen:
Just _**starting**_ to have a look at this, but looking at the screenshot I was wondering if it is worth having the breadcrumbs pointing to the prisoner profile? Similarly to what we have in Non-associations [for example](https://non-associations-dev.hmpps.service.justice.gov.uk/prisoner/A0772DZ/non-associations).

Otherwise you get to this page and there is no way out (well, without editing the URL directly really)

integration_tests/e2e/prisonerIncidentSummary.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+/** Build a closed report of the given type with crafted questions, all within the last 12 months. */
+function reportOfType(reference: string, type: string, questions: ReportWithDetails['questions']): ReportWithDetails {
+  const report = mockReport({
+    reportReference: reference,
+    reportDateAndTime: now,
+    type: type as ReportWithDetails['type'],
+    status: 'CLOSED',
+    location: andrew.prisonId,
+    withDetails: true,
+  }) as unknown as ReportWithDetails
```
xoen: Note that here the type assertion may possibly mask a potential bug or something you didn't want.

The `mockReport()` returns `DatesAsStrings<ReportWithDetails>` but you want `ReportWithDetails` (presumably?).

If that's the case you should convert the strings into dates, `convertReportWithDetailsDates()` does just that, possibly you could do something like:

```TypeScript
  const report = convertReportWithDetailsDates(
    mockReport({
      // ...
    })
  )
```

and that's why I'd advice against using type assertions (`as Something` as you're tricking TS to ignore potential issues) - only use very very very sporadically (if ever).

integration_tests/e2e/prisonerIncidentSummary.cy.ts
```diff
@@ -0,0 +1,67 @@
...
+import { andrew } from '../../server/data/testData/offenderSearch'
+import { mockReportingOfficer } from '../../server/data/testData/users'
+import { now } from '../../server/testutils/fakeClock'
+
+/** Build a closed report of the given type with crafted questions, all within the last 12 months. */
+function reportOfType(reference: string, type: string, questions: ReportWithDetails['questions']): ReportWithDetails {
+  const report = mockReport({
+    reportReference: reference,
+    reportDateAndTime: now,
+    type: type as ReportWithDetails['type'],
```
xoen: Another type assertion you should avoid.
The `type` parameter in the function signature should really be `Type` (unless I'm missing something).

This prevents accidentally passing something which is not an incident type and getting weird errors (or worse, not getting any), e.g. `'ASAULT_5'` would fly through even tho there is a typo.

Again, avoid type assertions (`as SomeType`) as much as possible (or ever) as TS then can't catch these problems.

server/middleware/requirePrisonerInCaseload.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * A denied prisoner is treated as Not Found rather than Forbidden: the user is legitimately
+ * signed in (so the app's 403-triggers-sign-out behaviour would be wrong), and we avoid
+ * confirming that a prisoner outside their caseload exists.
+ */
+export function requirePrisonerInCaseload(): RequestHandler {
+  return (_req, res, next): void => {
+    const { prisoner, user } = res.locals
+    if (!prisoner) {
+      next(new InternalServerError('requirePrisonerInCaseload() requires populatePrisoner() to run first'))
```
xoen: there is now the [`missingLocalsError()` utility function](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/6192ac0b554b0413ca250032af2b93a69245f9d2/server/errors.ts#L19) for this kind of (misconfiguration) scenario, so you should replace with something like:

```TypeScript
next(missingLocalsError('requirePrisonerInCaseload()', 'res.locals.prisoner'))
return
```

But you may have to add the entry for `res.locals.prisoner` to the mapping, e.g.:

```TypeScript
const middlewareMapping = {
  // ...
  'res.locals.prisoner': 'populatePrisoner()',
```

and you'd get a similarly helpful error along the line of

```Middleware configuration error: requirePrisonerInCaseload() requires res.locals.prisoner: populatePrisoner() was not executed first```

**Comment by Mjwillis** [comment]
> Just _**starting**_ to have a look at this, but looking at the screenshot I was wondering if it is worth having the breadcrumbs pointing to the prisoner profile? Similarly to what we have in Non-associations [for example](https://non-associations-dev.hmpps.service.justice.gov.uk/prisoner/A0772DZ/non-associations).
> 
> Otherwise you get to this page and there is no way out (well, without editing the URL directly really)

Yes I will be doing that - this is pass 1 of many, the next PR was going to add breadcrumbs

File: server/middleware/requirePrisonerInCaseload.ts (hunk 3 above)
Comment (Mjwillis) [reply server/middleware/requirePrisonerInCaseload.ts]: Good call — switched to \`missingLocalsError('requirePrisonerInCaseload()', 'res.locals.prisoner')\` and added the \`res.locals.prisoner\` → \`populatePrisoner()\` entry to the mapping. Fixed in 30968b5.

File: integration_tests/e2e/prisonerIncidentSummary.cy.ts (hunk 2 above)
Comment (Mjwillis) [reply integration_tests/e2e/prisonerIncidentSummary.cy.ts]: Agreed — the \`type\` parameter is now typed as \`Type\`, so the assertion is gone and a typo like \`ASAULT_5\` is a compile error. Fixed in 30968b5.

File: integration_tests/e2e/prisonerIncidentSummary.cy.ts (hunk 1 above)
Comment (Mjwillis) [reply integration_tests/e2e/prisonerIncidentSummary.cy.ts]: Thanks — replaced the assertion with \`convertReportWithDetailsDates(mockReport({ ... }))\`, which converts the date strings to \`Date\`s properly. Confirmed the stubs expect the \`Date\`-typed \`ReportWithDetails\`, so it lines up end-to-end. Fixed in 30968b5.

File: server/services/prisonerIncidentSummary/breakdowns.ts [hunk 4]
```diff
@@ -0,0 +1,239 @@
...
+  return questions.find(question => question.code === code)
+}
+
+/** Uppercase response *values* (the `response` field) given to a question, or [] if unanswered. */
+function responseValues(questions: Question[], code: string): string[] {
+  return findQuestion(questions, code)?.responses.map(response => response.response) ?? []
+}
+
+/** True if the question has any response whose *code* is in `responseCodes`. */
+function hasAnyResponseCode(questions: Question[], code: string, responseCodes: string[]): boolean {
```
xoen: I'd probably rename this and other similar parameters from `code` to `questionCode` to avoid ambiguity, especially given in some places like this we're dealing with both questions and answers (but in other places would still help avoid confusion).

server/services/prisonerIncidentSummary/breakdowns.ts
```diff
@@ -0,0 +1,239 @@
...
+//
+// Categories can be recorded two ways: a single-select "PLEASE SELECT CATEGORY OF FIND" (67187),
+// or "MULTIPLE TYPES" (67187 -> 218783) which walks each category as its own question. The rules
+// below detect a category across both paths.
+
+// "DESCRIBE THE DRUG FOUND" appears twice: single-select path (67190) and multiple-types path (67208).
+const DRUG_QUESTION_CODES = ['67190', '67208'] as const
+const NO_DRUG_VALUES = ['NONE FOUND', 'NIL'] as const
+
+function drugValues(questions: Question[]): string[] {
```
xoen: I think it would be clearer to use the word "response" instead of "value" in a lot of these places, e.g. `drugResponses()`, `NO_DRUG_RESPONSES`, `hasResponseNotIn()` (without the word "value"), `hasResponseStartingWith()`.

Possibly even `responseValues()` => `responses()` if it doesn't clash with anything else.

But I'd keep references to code as that's different (e.g. `hasAnyResponseCode()`).

xoen: Unless you think it could be ambiguous (between code vs. response)

server/views/pages/prisonerIncidentSummary.njk
```diff
@@ -0,0 +1,93 @@
...
+      <h1 class="govuk-heading-l">Incident summary</h1>
+
+      <p class="govuk-body">
+        Incidents involving this person in the 12 months since {{ summary.fromDate | longDate }}.
+        Draft, duplicate and not-reportable reports are excluded.
+      </p>
+
+      {% if summary.totalReports == 0 %}
+
+        <p class="govuk-body">No incidents have been recorded for this person in the last 12 months.</p>
```
xoen: Perhaps something like "This person hasn't been involved in any recorded incident in the last 12 months" (or something like that!) would be more accurate.

AFAICT the counts are based on the dates of the incident rather than the date where the incident was recorded.

server/services/prisonerIncidentSummary/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  let totalPages = 1
+  do {
+    // eslint-disable-next-line no-await-in-loop -- pages must be fetched sequentially
+    const response = await api.getReports({
+      involvingPrisonerNumber: prisonerNumber,
+      incidentDateFrom: fromDate,
+      status: INCLUDED_STATUSES,
+      page,
+      size: PAGE_SIZE,
+      sort: ['incidentDateAndTime,DESC'],
```
xoen: I don't imagine the order matters much in this case.
In fact I think `getReports()` already defaults to `incidentDateAndTime,DESC` and you should be able to remove this line AFAICT.

server/services/prisonerIncidentSummary/index.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * Fetch the questions of every report whose type matches the given active version, in parallel.
+ * Older type versions are not understood by the detail extractors so they are excluded.
+ */
+async function activeVersionQuestions(
+  api: IncidentReportingApi,
+  reports: ReportBasic[],
+  activeType: string,
+): Promise<Array<Awaited<ReturnType<IncidentReportingApi['getReportWithDetailsById']>>['questions']>> {
```
xoen: This is a mouthful...isn't this really equivalent to `Promise<Question[][]>`?

(aside from not dynamically be in sync with the return type of `getReportWithDetailsById()` which I don't think it's worth the obfuscation, especially as if that ever changes you'd get a type error here anyway)

server/services/prisonerIncidentSummary/index.ts
```diff
@@ -0,0 +1,136 @@
...
+  const matching = reports.filter(report => report.type === activeType)
+  const detailed = await Promise.all(matching.map(report => api.getReportWithDetailsById(report.id)))
```
xoen, line 101: This is worth checking in `preprod` against a prisoner with a lot of reports...if I'm reading this correctly it would hammer the API with a lot of concurrent requests to get all the reports for the given prisoner at once.

It's a tricky one because it may work for most small/medium numbers of reports but then you get a big list of reports and it would fail.

I don't know if it's worth batching them up from the get go to avoid this situation or if it's almost premature optimisation 🤔 

May be worth finding a prisoner with a huge amount of incidents and see if that play ball in `preprod`.

server/views/pages/prisonerIncidentSummary.njk
```diff
@@ -0,0 +1,93 @@
...
+        {{ govukTable({
+          caption: "All incidents by type",
+          captionClasses: "govuk-table__caption--m",
+          firstCellIsHeader: true,
+          head: [
+            { text: "Incident type" },
+            { text: "Number of reports", format: "numeric" }
+          ],
+          rows: overallRows
+        }) }}
```
xoen: Maybe there are other differences I'm not spotting, but isn't this pretty much the `breakdownTable()` macro with the exception of the `Detail` header being `Incident type`?

Maybe you could add another parameter to that macro for the label for the first column?

server/services/prisonerIncidentSummary/index.ts
Mjwillis: Yes will try it in preprod with a large number

server/services/prisonerIncidentSummary/breakdowns.ts
Mjwillis: Good shout — renamed the question-code parameters to `questionCode` across the helpers so they read unambiguously against the response codes. Fixed in 9ce0be7.

server/services/prisonerIncidentSummary/breakdowns.ts
Mjwillis: Agreed — switched the response-text helpers/constants to the "response" wording (`responses()`, `drugResponses()`, `NO_DRUG_RESPONSES`, `hasResponseNotIn()`, `hasResponseStartingWith()`, plus the `*_RESPONSES` drug constants) and kept the code-based names as-is (`hasAnyResponseCode`, `responseCodes`). Fixed in 9ce0be7.

server/views/pages/prisonerIncidentSummary.njk
Mjwillis: Good point — the window is by incident date, so reworded the empty state to "This person hasn't been involved in any recorded incident in the last 12 months." Fixed in 9ce0be7.

server/services/prisonerIncidentSummary/index.ts
Mjwillis: Removed — ordering doesn't matter here as we only group and count, so dropped the explicit `sort`. Fixed in 9ce0be7.

server/services/prisonerIncidentSummary/index.ts
Mjwillis: Agreed it was a mouthful — simplified to `Promise<Question[][]>` (importing `Question`). If `getReportWithDetailsById()` ever changes shape we'd still get a type error at the call site. Fixed in 9ce0be7.

server/views/pages/prisonerIncidentSummary.njk
Mjwillis: Nice — the only real difference was the first-column header, so added a `firstColumnHeader` parameter (defaulting to "Detail") to `breakdownTable` and reused it for the overall table with "Incident type", dropping the duplicated `govukTable` block. Fixed in 9ce0be7.

xoen (approved):
Nice work ☺️

PR #871, opened 2026-06-12
IR-1681 Add breadcrumb to prisoner incident summary page

Mjwillis:
## Summary
Adds a breadcrumb to the prisoner incident summary page (`/prisoner/:prisonerNumber/incident-summary`) so users can navigate back to the prisoner's profile in Digital Prison Services.

The trail mirrors the Incentives service: **Digital Prison Services > {surname, forename}**, where the name links to `{dpsUrl}/prisoner/{prisonerNumber}`.

This is step 1 of linking the prisoner profile to the incident summary (the profile-side link is a separate PR in hmpps-prisoner-profile under the same ticket).

## Changes
- `prisonerIncidentSummary.njk` — add a `containerStart` block rendering `govukBreadcrumbs`, using the existing app pattern.
- `utils.ts` / `nunjucksSetup.ts` — add `reversedNameOfPerson` helper + Nunjucks filter for the surname-first name format.
- Unit tests for the helper and the breadcrumb render.

## Testing
- `npm run typecheck` ✅
- `npm run lint` ✅
- `npx jest utils prisonerIncidentSummary` ✅ (332 passing)

### DIFF
--- a/server/utils/utils.ts
+++ b/server/utils/utils.ts
@@ -32,6 +32,18 @@ export function convertToTitleCase(sentence: string | undefined | null): string 
 export const nameOfPerson = (prisoner: { firstName: string; lastName: string }): string =>
   `${convertToTitleCase(prisoner.firstName)} ${convertToTitleCase(prisoner.lastName)}`.trim()
 
+/**
+ * Surname-first display form of a person’s name, matching the Digital Prison Services convention
+ * used in breadcrumbs (e.g. linking back to a prisoner’s profile).
+ * { "firstName": "DAVID", "lastName": "JONES", … } → "Jones, David"
+ */
+export const reversedNameOfPerson = (prisoner: { firstName: string; lastName: string }): string => {
+  const firstName = convertToTitleCase(prisoner.firstName)
+  const lastName = convertToTitleCase(prisoner.lastName)
+  if (firstName && lastName) return `${lastName}, ${firstName}`
+  return `${lastName}${firstName}`.trim()
+}
+
 export const initialiseName = (fullName?: string): string | null => {
   // this check is for the authError page
   if (!fullName) return null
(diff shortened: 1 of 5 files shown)


server/utils/utils.ts
```diff
@@ -32,6 +32,18 @@ export function convertToTitleCase(sentence: string | undefined | null): string
 export const nameOfPerson = (prisoner: { firstName: string; lastName: string }): string =>
   `${convertToTitleCase(prisoner.firstName)} ${convertToTitleCase(prisoner.lastName)}`.trim()
 
+/**
+ * Surname-first display form of a person’s name, matching the Digital Prison Services convention
+ * used in breadcrumbs (e.g. linking back to a prisoner’s profile).
+ * { "firstName": "DAVID", "lastName": "JONES", … } → "Jones, David"
+ */
+export const reversedNameOfPerson = (prisoner: { firstName: string; lastName: string }): string => {
```
xoen, line 40: Do we need to add this back? (I removed it in [this PR](https://github.com/ministryofjustice/hmpps-incident-reporting/pull/857/changes#diff-4d93e707b3f1f1722d2721d2410b8b42ee601cbe10942a90830419f23eecda57L24) as it was unused)

You could just use `nameOfPerson` like in the existing pages?

Mjwillis: Good catch that this is a re-add. I brought it back deliberately for the breadcrumb here: it links back to the prisoner's DPS profile and follows the DPS / Incentives breadcrumb convention, which shows the name surname-first (e.g. `Brooks, Kyle` on the Incentives "incentive details" page). `nameOfPerson` gives forename-first (`Kyle Brooks`), which would be inconsistent with that pattern.

So it's now actually used — by this breadcrumb and by the IR-1683 incident-list drill-down (stacked in #873) — rather than the dead code it was when you removed it in #857. Happy to switch to `nameOfPerson` if you'd rather keep consistency with the in-service pages over matching the DPS breadcrumb style.

xoen (approved):

PR #872, opened 2026-06-12
IR-1644 legacy type labels

jonesj2193:
Primary goal for this PR is to change the labels relating to inactive types in the autocomplete filter for incident types on the dashboard. The changes were to add test to the label to indicate the type is inactive with the text following the format "(inactive since MMM YYYY)". It was also to group all inactive types together at the bottom of the presented list. 

This filter actually uses the type family, as we display all types of the same family with the same label i.e. ASSAULT_1, ASSAULT_2 etc are all displayed as "Assault" and the filter includes all types belonging to that family. So for this work, it was not simply to check if a type is still active or not, but to check if all types within a family are inactive, and if they were, use the most recent date if there are multiple expiry dates present within a family. 

In order to do this, the following functions were created:

- `areTypeFamiliesInactive() `- to determine whether a type family is now inactive (i.e. do all types within the family have an "activeTo" value that is in the past. 
- `getTypeFamilyExpiryDates()` - to get the most recent "activeTo" date associated with the family type ready to put in to the label on the autocomplete filter. 

In the dashboard controller, these functions are used to filter out, edit labels, and rearrange the filter items accordingly. 

Additional work was required along the way to refactor some of the types relating to incident types and families. 

Further work on refactoring these, along with combining/refactoring some of the configs relating to types and families would help maintainability in future as this information is now spread and in some cases duplicated across several files.


Mjwillis (dismissed):
LGTM

server/reportConfiguration/constants/typeActivePeriods.ts
```diff
@@ -101,3 +102,53 @@ export function isTypeActive(code: string, at: Date = effectiveNow()): boolean {
...
+export function areTypeFamiliesInactive(
+  typeDetails: IncidentType[],
+  at: Date = effectiveNow(),
+): Record<FamilyCode, boolean> {
+  return typeDetails.reduce(
+    (acc, item) => {
+      acc[item.familyCode] = (acc[item.familyCode] ?? true) && !isTypeActive(item.code, at)
+      return acc
+    },
+    {} as Record<FamilyCode, boolean>,
```
xoen: You don't need this type assertion (they're generally bad), you can annotate the closure with types and it would infer the correct return type for you, e.g.:

```TypeScript
return typeDetails.reduce(
  (acc: Record<FamilyCode, boolean>, item: IncidentType) => {
    // ...
```

(and if in the future if for whatever reason the logic changes it wouldn't accidentally return a value of the wrong type)

server/reportConfiguration/constants/typeActivePeriods.ts
```diff
@@ -101,3 +102,53 @@ export function isTypeActive(code: string, at: Date = effectiveNow()): boolean {
...
+      const latestDate = expiryDates.length > 0 ? new Date(Math.max(...expiryDates.map(d => d.getTime()))) : null
+
+      return [
+        familyCode,
+        latestDate
+          ? `${new Intl.DateTimeFormat('en-US', { month: 'short' }).format(latestDate)} ${latestDate.getFullYear()}`
+          : null,
+      ]
+    }),
+  ) as Record<FamilyCode, string | null>
```
xoen: You don't need this type assertion AFAICT. I've tried to remove it locally (with `strict: true`) and it didn't complain at all.

server/reportConfiguration/constants/types.ts
```diff
@@ -1,7 +1,15 @@
 // Generated with ./scripts/importDpsConstants.ts at 2025-08-19T16:38:54.769Z
 
+export type IncidentType = {
```
xoen: This file is autogenerated with the `./scripts/importDpsConstants.ts` script so we shouldn't be touching it directly.
E.g. the script should be updated, otherwise next time these files are re-generated this change will be lost.

I guess the question is, do we need this type? (the `types` constant is inferring the type fine) or could this type be defined somewhere else perhaps (non-autogenerated)

server/reportConfiguration/constants/typeFamilies.ts
```diff
@@ -1,7 +1,13 @@
 // Generated with ./scripts/importDpsConstants.ts at 2025-08-19T16:38:55.854Z
+import { FamilyCode } from './types'
+
+export type IncidentTypeFamily = {
```
xoen: (same point as in different comment, we can't change the generated files directly)

AFAICT, the only place where this type is really used is in the signature of `getTypeFamilyExpiryDates()`. So you could move this type in that module.

(in other places TS is happy to infer the type for you I think, without the type annotation)

jonesj2193: This file has been changed 7 times since that date including once by yourself. So I didn't think it was an issue anymore as we no longer sync this back to or from NOMIS anyway.

server/reportConfiguration/constants/types.ts
jonesj2193: Similar to my other reply. This one has been changed 11 times since that auto generated comment.

server/reportConfiguration/constants/typeFamilies.ts
xoen: The only changes I'm seeing are to the data itself - e.g. as the incident types evolved (to align with changes on the API side) - not to the actual format (which would require a change to the script that generates it - I'd personally avoid going into that rabbit hole).

jonesj2193: But then if we haven't run it for 10 months, and all the changes are being done manually, then is there a need to run the script again? 
Also with the proposed refactoring of all of this information, it may render this files/scripts as they currently are redundant anyway. 
Also the primary reason I added it in was to help with typing more broadly as they are all just inferred all the time from what is imported in that structure, rather than actually setting what we expect to see somewhere and then it matching. When we are wanting to be stricter on types, this feels like a better approach otherwise the cart is pulling the horse a bit.

xoen: Just to be clear, those "manual" changes are just the same changes that the script would have generated to reflect the changes to the API side. E.g. if we ran the script it would generate the same values and types.

The reason why the type for these values is not set explicitly is mainly because the value is coming from the API and the generation script is generating different values for different constants. Also TypeScript is decent at inferring types so in this case it would be redundant.

(e.g. `let x = 'foo'` is generally preferred to `let x: string = 'foo'` - although this is simple example of course, the uniform nature of the values returned from the API for a certain constant makes the type annotation superfluous)

I would leave these generated files alone and either see if these new types are actually needed or if they really are put them somewhere else. I'm happy to catch up about this if it's easier.

jonesj2193: Well 2 points, firstly, if they are to match the API side, shouldn't that drive the process? Shouldn't we make the changes there and then run the script to update the values? my point is that the script hasn't been run for 10 months, so there could very well be (and maybe likely is) differences now between the API and what's in the front end. So if we aren't running the script to drive this file, then do we need the script anymore? 
And secondly then, if we aren't regularly using the script to generate it, and we have mentioned refactoring these configs shortly, that will render the existing structure and script redundant then anyway. As we would then either need to rewrite that script to make whatever we decide we want going forward, or just remove it and maintain it manually, which is what has actually been happening for the last 10 months anyway

jonesj2193: Also further to my point, I am not the first person to make "breaking changes" to [this file](https://github.com/ministryofjustice/hmpps-incident-reporting/commit/a60d023f736b2885c4a4f824ceb44ad8d9b85cb2)

jonesj2193: Which was part of a PR you reviewed and approved. Just highlighting that there wasn't a problem then, so why now?

xoen: Regarding that PR, I guess that sneaked through :| (it's bound to happen given the increase in volume and size of some of these PRs, plus we are only humans) - regardless, my point about not editing a generated file manually stands (but if, as you say, we stop generating these 2 files that would be different going forward).

@Mjwillis what do you think? Do you have a view on whether we should continue generate types/type families values/types off the data exposed by the API? (is the data in the API exposed because it is used somewhere else? I can't quite remember)

Or should stop doing that and just maintain these 2 files manually going forward instead (but that would mean update the `scripts/importDpsConstants.ts` script to remove the entries for `types`/`typeFamilies`, assuming that's the consensus)

xoen (approved):

PR #875, opened 2026-06-15
IR-1664 food refusal labels

jonesj2193:
Updated the food refusal labels for the question `Which type of refusal occurred?` to include the timing window that is relevant.


server/reportConfiguration/types/FOOD_REFUSAL_2.ts
```diff
@@ -41,7 +41,7 @@ export const FOOD_REFUSAL_2: IncidentTypeConfiguration = {
           code: '950003',
           response: 'Food and liquid',
           active: true,
-          label: 'Food and liquid',
+          label: 'Food and liquid (for 24 hours or more)',
```
xoen: Isn't the label 'Liquid and food (for 24 hours or more)' instead?

(but I may be wrong as that ticket has been a bit confusing with the comments, other changes, etc...tbh)

jonesj2193: I didn't notice it had swapped around tbh. Pushing the swapped version now.

xoen (approved):

PR #877, opened 2026-06-15
IR-1646: Stricter TypeScript (Part 2)

xoen:
Fix more of the type errors raised when `"strictNullChecks": true`.

NOTE: There are a few more but this is probably big enough to be reviewed now. I could create another PR for the outstanding fixes to the remaining type errors.


Mjwillis (approved):
LGTM

PR #880, opened 2026-06-17
IR-1696 Hide incident report link when user cannot view the report

Mjwillis:
## Why

On the prisoner incident list (`/prisoner/:prisonerNumber/incident-summary/incidents`, added in IR-1683), each row's "Incident ID" cell rendered a hyperlink to `/reports/:reportId` **unconditionally**. That destination route is gated by the canonical VIEW permission check (`permissions.allowedActionsOnReport(report).has('VIEW')`), which requires a service role **and** the report's establishment in the user's caseloads (or PECS access). Because a prisoner's 12-month list can include incidents at establishments outside the current user's caseloads, a user could click a link that immediately logged them out.

## What

Render the hyperlink only when the user can actually view that report; otherwise show the report reference as plain text. No new permission logic — reuses the single source of truth in `rulesClass.ts`.

- **`services/prisonerIncidentSummary/list.ts`** — add `canView: boolean` to `PrisonerIncidentListRow`; thread a `Permissions` instance through `getPrisonerIncidentList` → `rowsForReport`; compute `canView` once per report from the raw `status`/`location` via `permissions.allowedActionsOnReport({ status, location }).has('VIEW')`.
- **`routes/prisonerIncidentSummary/index.ts`** — pass `res.locals.permissions` into the service.
- **`views/pages/prisonerIncidentList.njk`** — emit the `<a>` only when `row.canView`, else plain `reportReference`.

## Tests

- `list.test.ts` — threads a reporting-officer `Permissions` instance through existing calls; adds an in/out-of-caseload assertion (MDI → `canView` true, LEI → false).
- `index.test.ts` — existing `/incidents` test now asserts the viewable report is linked; new test asserts an out-of-caseload report shows the reference but no `/reports/` anchor.

### DIFF
--- a/server/services/prisonerIncidentSummary/list.ts
+++ b/server/services/prisonerIncidentSummary/list.ts
@@ -2,6 +2,7 @@ import type {
   IncidentReportingApi,
   PrisonerInvolvement,
   Question,
+  ReportBasic,
   ReportWithDetails,
 } from '../../data/incidentReportingApi'
 import type { PrisonApi } from '../../data/prisonApi'
@@ -11,7 +12,6 @@ import {
   getTypeDetails,
   prisonerInvolvementRolesDescriptions,
   statusesDescriptions,
-  type Status,
 } from '../../reportConfiguration/constants'
 import { ACTIVE_DETAIL_TYPES } from './breakdowns'
 import { fetchAllReports, twelveMonthsAgo } from './prisonerReports'
@@ -47,9 +47,12 @@ export interface PrisonerIncidentListRow {
   location?: string
   establishment: string
   status: string
-  /** Raw report status/location, used by the template's VIEW check to gate the link to /reports/:id. */
-  reportStatus: Status
-  reportLocation: string
+  /**
+   * Raw report status/location — used by the template's VIEW check to gate the link to /reports/:id.
+   * NB the row's own `status`/`location` above are display-translated / free-text and must NOT be
+   * used for that check; the establishment code lives here in `report.location`.
+   */
+  report: Pick<ReportBasic, 'status' | 'location'>
 }
 
 export interface PrisonerIncidentList {
@@ -203,8 +206,7 @@ function rowsForReport(
     status: statusesDescriptions[report.status] ?? report.status,
     // Raw values (NOT the display-translated status/location above) so the template's VIEW check
     // matches the gate enforced by the /reports/:id route.
-    reportStatus: report.status,
-    reportLocation: report.location,
+    report: { status: report.status, location: report.location },
   }))
 }
 
--- a/server/services/prisonerIncidentSummary/list.ts
+++ b/server/services/prisonerIncidentSummary/list.ts
@@ -5,6 +5,7 @@ import type {
   ReportWithDetails,
 } from '../../data/incidentReportingApi'
 import type { PrisonApi } from '../../data/prisonApi'
+import type { Permissions } from '../../middleware/permissions'
 import { AgencyType } from '../../data/constants'
 import { isPecsRegionCode } from '../../data/pecsRegions'
 import {
@@ -46,6 +47,8 @@ export interface PrisonerIncidentListRow {
   location?: string
   establishment: string
   status: string
+  /** Whether the current user is permitted to view this report (gates the link to /reports/:id). */
+  canView: boolean
 }
 
 export interface PrisonerIncidentList {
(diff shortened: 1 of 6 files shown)


server/services/prisonerIncidentSummary/list.ts
```diff
@@ -46,6 +47,8 @@ export interface PrisonerIncidentListRow {
   location?: string
   establishment: string
   status: string
+  /** Whether the current user is permitted to view this report (gates the link to /reports/:id). */
+  canView: boolean
```
xoen: I don't think this is the right place for permission checking.

This module concern is to do with retrieval/wrangling of the data (what it was doing before) - while the permission checking is a presentation concern.

I think it would be better to leave everything as it was and add a Nunjucks helper that you then use in the template, e.g. something _along the lines_ of: 

```TypeScript
// define helper somewhere
function canViewReport(report: Pick<ReportBasic, 'status' | 'location'>, permissions: Permissions): boolean {
  return permissions.allowedActionsOnReport({ status: report.status, location: report.location }).has('VIEW')
}

// register with Nunjucks
njkEnv.addFilter('canViewReport', canViewReport)

// then use it in the template
{% if canViewReport(report, permissions) %}
  // ...
```

I think then the rest could be roughly stay as before but I may be wrong.

xoen (changes requested):
See comment on moving the "can view" logic just into Nunjucks/template.

server/services/prisonerIncidentSummary/list.ts
Mjwillis: Agreed the VIEW check is a presentation concern, not data wrangling. I've moved it into a canViewReport(reportLike, permissions) Nunjucks helper and the template now uses it directly. 

The service-layer threading of Permissions is gone, and the route change is reverted

One thing was that the template iterates over the list rows, not the report objects, and the row's status is display-translated while location is the in-establishment free-text answer (not the establishment code). The VIEW check needs the raw status/location, so the row now carries reportStatus/reportLocation for the helper to read.

 That keeps the permission decision in the template while the service only exposes report data — which I think is better.

Mjwillis:
> See comment on moving the "can view" logic just into Nunjucks/template.
I've made your suggested changes

server/services/prisonerIncidentSummary/list.ts
```diff
@@ -204,7 +201,10 @@ function rowsForReport(
     location: location(report),
     establishment,
     status: statusesDescriptions[report.status] ?? report.status,
-    canView,
+    // Raw values (NOT the display-translated status/location above) so the template's VIEW check
+    // matches the gate enforced by the /reports/:id route.
+    reportStatus: report.status,
```
xoen: Do you need these 2 properties? isn't `canViewReport()` happy with any object that has `status`/`location`? (like row in this case) 🤔

server/views/pages/prisonerIncidentList.njk
```diff
@@ -49,7 +49,7 @@
         {% set rows = [] %}
         {% for row in prisonerIncidentList.rows %}
           {% set linkHtml %}
-            {% if row.canView %}
+            {% if canViewReport({ status: row.reportStatus, location: row.reportLocation }, permissions) %}
```
xoen: Couldn't you just call `canViewReport()` on the row (as it has `status`/`location` and "quacks" like a report)?

server/services/prisonerIncidentSummary/list.ts
xoen: Because when I commented earlier I thought this module would effectively be untouched and `canViewReport()` can be used against these rows as-they-are (or were anyway).

Mjwillis: Good shout on the shape — I've nested the raw values as `row.report` so it now reads `canViewReport(row.report, permissions)`. I can't use the row's own `status`/`location` though: by the time they hit the template they're already display-translated (`"Awaiting review"`) and the free-text in-cell location (`"Gym"`) respectively, whereas the VIEW check needs the raw status enum and the establishment code for the caseload check — using the row's values would hide every link. So `row.report` carries the raw `status`/`location` for that purpose.

xoen: > I can't use the row's own status/location though: by the time they hit the template they're already display-translated ("Awaiting review") and the free-text in-cell location ("Gym") respectively, whereas the VIEW check needs the raw status enum and the establishment code for the caseload check

I see, these properties contained processed values but it makes sense as they were rendered, sorry I missed that!

xoen (approved):

xoen (approved):

server/views/pages/prisonerIncidentList.njk
xoen: > Couldn't you just call canViewReport() on the row (as it has status/location and "quacks" like a report)?

Answered in other comment: No, the values in the `status`/`location` properties on the row have processed values not the ones needed to check permission (but these values are now under `row.report` and used here).

PR #881, opened 2026-06-17
IR-1646: Stricter type checking (Part 3)

xoen:
Enable `strict` and resolve or document the last few type errors.
- backport some of the type improvements from [this TypeScript template PR](https://github.com/ministryofjustice/hmpps-template-typescript/pull/718).
- document 3 type errors related to `strictFunctionTypes` (with `// @ts-expect-error`) so that we can enable `strict` checking
- improve various types and function signatures to reflect reality
- better handling of `null` and `undefined` values (e.g. in `server/data/incidentTypeConfiguration/formWizard.ts`)
- sporadic use of type/null assertions in few cases where I couldn't figure out an alternative (but documented)
  - for example the `DatesAsString<T>` type seem to widen the type of the fields turning something like `ApiUserType` (or similar union types) into plain `string`. Maybe there is a way of improving this but for now I've worked around these few instances and document it for future us
- in tests I've been slightly more liberal with the use of `// @ts-expect-error` as often the input is a crude version of the real thing or it's malformed by design (to test edge cases)
- integration tests' `tsconfig.json` - which is extended from base config file - to be stricter than before but the `strictNullChecks` and `strictFunctionTypes` options are currently turned off as there are still quite a few type errors in integration tests related to this. Fixing these is probably be left for another time, but at least the overall type checking is stricter than it was before.


jonesj2193 (approved):
LGTM

PR #890, opened 2026-07-01
IR-1733: Fix bug list of active types not updated

xoen:
Incident types can have a date range when they'll come online or be deactivated. The theory is that, on the day new types versions come online, we don't need to do anything and the service will show the correct list of incident types.

However this morning we discovered that the list of active types was not updated until the pods were restarted.

This was caused by the `typeActivePeriods` module exporting the computed list. This means that the list was a "snapshot" created at the point in time when the module was first imported (aka shorly after the pods started).

The smallest possible fix to this problem is to remove this calculated list from the module and calculate the updated list inside the request handler using it.

### DIFF
--- a/server/reportConfiguration/constants/typeActivePeriods.ts
+++ b/server/reportConfiguration/constants/typeActivePeriods.ts
@@ -125,8 +125,6 @@ export function areTypeFamiliesInactive(
   )
 }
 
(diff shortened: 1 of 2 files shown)


Mjwillis (approved):
But does this fix the incident types becoming active too?

PR #892, opened 2026-07-02
IR-1733: Types list up-to-date with retired/activated types

xoen:
The list of active types was not updated when types were retired or activated and required the pods to be restarted to pick-up this change.

This was caused by the fact that the FormWizard field for the `type` was "static" in a module and not changing once the module was imported the first time.

To ensure this list is always up-to-date the field's items are updated in a request handler.


jonesj2193 (approved):
Looks to make sense to me. Hard to test when it would next work though I guess.

PR #896, opened 2026-07-09
IR-1762: Dashboard refactoring

xoen:
The dashboard request handler was very long (250+ lines), difficult to read/update.
A lot of the logic to read values from query params/session was mixed with logic to validate or convert these values into an higher level format.

This refactoring moved most of the logic out of the request handler.

Key concepts:
- `UiFilters` is what you get from the UI (e.g. the mostly raw strings)
- `Filters` is the high level - processed and cleaned - filters, mainly passed to `getReports()` to get the actual reports
- reading of UI filters from query/session happens in `readUiFilters()` (which also apply defaults for `sort`/`order`/`incidentStatuses`)
- validation is also moved out of the request handler into `validateUiFilters()`
- and transformation into the high-level filters happens in `filtersFromUiFilters()`
- I've also simplified some of the types by turning `Thing | Thing[]` into just `Thing[]` (normalised early on. by `readUiFilters()` so that subsequent code doesn't have to check if it's an array or not at every step

Structure:
- `validateUiFilters()` relies on a bunch of helper functions like `validateSearchId()`, `validateLatestUserActions()`, etc...
  - there are also smaller utility functions to check validity of individual fields, e.g. `validPrisonerNumber()`, `validTypeFamily()`, etc...
- `filtersFromUiFilters()` also relies on a bunch of helper functions that convert lower level-values into higher-level filter values, e.g. `processLocation()`, `processStatus()`, etc...
- the idea is that these concepts are behind nice names that makes higher-level logic easier to read and follow and more intent-revealing

Other changes:
- Introduced/tweak some of the type names
- `Filters` fields were renamed so that they match the filters properties used to get the list of reports
- About the `page` property
  - in `UiFilters` is a optional string and start from 1 as this is the value coming from the UI where pagination starts at 1
  - in `Filters` the page is a mandatory number and starts from 0 as this is the value used to get filtered reports from the API

Ultimately the goal is not perfection but rather a simplification of the request handler and moving the filters handling in its own module. There are probably more improvements that can be made.

Also note that the aim was not remove code but rather give it structure and improve readability and make it less brittle. Most of the logic is still there, just in different places. In fact potentially there are _more_ line of code now because of how some of the logic is extracted into smaller, well-named, self-contained functions.

### DIFF
--- a/server/routes/dashboard/filters.ts
+++ b/server/routes/dashboard/filters.ts
@@ -17,7 +17,7 @@ import { parseDateInput } from '../../utils/parseDateTime'
 import format from '../../utils/format'
 import { type Order } from '../../data/offenderSearchApi'
 import { hasInvalidValues } from '../../utils/utils'
-import { type ApiUserAction, apiUserActions } from '../../middleware/permissions'
+import { type ApiUserAction, type Permissions, type UserAction, apiUserActions } from '../../middleware/permissions'
 
 // `latestUserActions` can include 'REQUEST_REMOVAL' (not a valid `ApiUserAction`)
 // which maps/is replaced with 'REQUEST_NOT_REPORTABLE' and 'REQUEST_DUPLICATE'
@@ -46,6 +46,7 @@ interface Filters {
   toDate?: Date
   type?: Type[]
   status?: Status[]
+  userAction?: ApiUserAction[]
   sort: string[]
   page: number
 }
@@ -143,7 +144,7 @@ export function validateUiFilters(uiFilters: UiFilters, useWorklists: boolean): 
   return errors
 }
 
-export function filtersFromUiFilters(uiFilters: UiFilters, useWorklists: boolean): Filters {
+export function filtersFromUiFilters(uiFilters: UiFilters, useWorklists: boolean, permissions: Permissions): Filters {
   let page = (uiFilters.page && parseInt(uiFilters.page, 10)) || 1
   if (page < 1) {
     page = 1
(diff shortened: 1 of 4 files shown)


server/routes/dashboard/filters.ts
```diff
@@ -0,0 +1,457 @@
...
+  validateLocation({ uiFilters, errors, permissions, userCaseloadIds })
+  validateDateRanges(uiFilters, errors)
+  validateIncidentStatuses(uiFilters, errors, useWorkLists)
+  validateLatestUserActions(uiFilters, errors)
+  validateTypeFamily(uiFilters, errors)
+
+  return errors
+}
+
+export function filtersFromUiFilters({
```
jonesj2193, line 165: Maybe this could just be renamed to something a little less confusing? Thinking of somebody coming in who has no idea what is going on. Maybe something either along the lines of `filtersFromUiInputs` / `filtersFromUiQuery` or just something to indicate where these are then used, like `ApiFiltersFromUiFilters`. 
Just to help make it a little more explicit of the transition between the two.

xoen: The rational behind the name was that there is a distinction between the filters coming from the UI (`UiFilters`) and the filters we then use to get the reports (just `Filters`).

`readUiFilters()` returns `UiFilters`, while `filtersFromUiFilters()` take an instance of `UiFilters` and returns `Filters`, hence the name.

(I also took inspiration from some of the function names in Adjudications)

As discussed offline, I'll leave this name (or at least for now).

jonesj2193 (approved):
Happy to merge in when you're ready

PR #899, opened 2026-07-13
IR-1770 Add admin screen to sync incident type config into NOMIS

Mjwillis:
## What & why

DPS is the source of truth for incident-type questionnaires, but NOMIS still needs the same config so reports raised in this service sync back correctly (the data feeds reporting and the data-science teams), even though NOMIS screens are switched off everywhere.

Until now this was done by running `scripts/buildNomisIncidentTypePayload.ts` locally and calling Prison API by hand with a **prod token** — which almost no developer is allowed to hold. This PR replaces that with a **secret, admin-gated screen** at `/admin/sync-nomis` and deletes the script.

## What it does

`/admin/sync-nomis` (no homepage link; gated on `INCIDENT_REPORTS__ADMIN`, held in addition to a base role):

1. **Select** an incident type that is **active now or due to go live**. Upcoming versions are listed with their go-live date so a new version can be prepared in NOMIS ahead of the switch-over (this also tells apart two versions that share a description, e.g. Tool loss v1/v2).
2. **Confirm** — the page checks NOMIS and says whether the type will be **created** (not there yet) or **updated** (already there), with a short summary (including a *Goes live* row for upcoming types).
3. **Result** — performs the create/update, then reports whether NOMIS now matches the config this service holds (or lists the differences).

Screenshot 2026-07-14 at 07 43 02

Screenshot 2026-07-14 at 07 43 16

## Pre-loading a version before it goes live

A new version (e.g. `TOOL_LOSS_2`, live 1 Aug) must be in NOMIS **before** the switch-over so the data is ready to use on day one. The list therefore includes upcoming types (via a new `isTypeActiveOrUpcoming` predicate, alongside the existing `isTypeActive`), and they are written with `active: true` — in Prison API that flag only sets the expiry date and list ordering, it does not gate reads, and NOMIS is not used to create reports. So one pre-sync leaves the data ready with **no action needed on the day**. Retiring the previous version in NOMIS on its end date is out of scope (a separate future task).

## Changes

- **Relocated the payload builder** out of the deleted script into shared code (`server/data/incidentTypeConfiguration/nomisPayload.ts`), preserving the DPS→NOMIS field renames (`commentRequested→commentRequired`, `dateMandatory→dateRequired`, `onlyOneAllowed→singleRole`) and the flow-based question ordering.
- **Prison API client** (`server/data/prisonApi.ts`): added `createIncidentTypeConfiguration`, `updateIncidentTypeConfiguration`, and an `incidentTypeConfigurationExists` helper (GET → 404 means "create").
- **Post-sync comparison** (`server/data/incidentTypeConfiguration/nomisCompare.ts`): normalises the request we sent and the NOMIS response (different field names) and diffs them with `util.isDeepStrictEqual`.
- **Upcoming types** (`server/reportConfiguration/constants/typeActivePeriods.ts`): `isTypeActiveOrUpcoming` and `upcomingActivationDate` drive the broadened list and the go-live labelling.
- **Permissions**: added `isAdmin`; new `/admin` router subtree gated on it.
- **Views**: `select` / `confirm` / `result` pages.
- **Deleted** `scripts/buildNomisIncidentTypePayload.ts`; documented the screen in the README.

## Prison API role

The write relies on the service's **system client** holding the Prison API role **`PRISON_API__INCIDENT_TYPE_CONFIGURATION_RW`**, which is now granted in **every environment**. A failure from Prison API is therefore treated as a genuine error and left to the standard error handler (no special "missing role" screen). The Prison API endpoints already exist, so no prison-api change is required.

## Testing

- Unit tests for the payload builder, the comparison util, and the new `isTypeActiveOrUpcoming` / `upcomingActivationDate` helpers (including the switch-over boundary).
- Route tests: admin gating (non-admin is signed out), select → confirm (create vs update), the create and update paths, an upstream-failure propagating to the error handler, and the upcoming-type flow (offered with its go-live date, selectable, and the *Goes live* row on confirm; retired versions drop off). These render the real Nunjucks views.
- A Cypress integration test drives the screen end-to-end through the running app against a WireMock-stubbed Prison API.
- `typecheck`, `lint`, and the full `npm test` suite pass locally.


server/data/incidentTypeConfiguration/nomisPayload.ts
```diff
@@ -0,0 +1,126 @@
...
+  const dfs = (code: string | null | undefined): void => {
+    if (!code) return
+    const question = questionsByCode[code]
+    if (!question || visited.has(code)) return
+    visited.add(code)
+    orderedCodes.push(code)
+
+    const seenNext = new Set<string>()
+    question.answers.forEach(answer => {
+      if (!answer.active) return
```
xoen, line 66: You filter out inactive answers, do you also want to filter out inactive questions? (AFAICT currently all questions are considered)

server/data/incidentTypeConfiguration/nomisPayload.ts
```diff
@@ -0,0 +1,126 @@
...
+  /** NOMIS incident-type code (max 12 chars) — distinct from the DPS type code */
+  nomisCode: string
+  /** Incident type description */
+  description: string
+  /** Whether the type is active */
+  active: boolean
+}
+
+/**
+ * Orders questions the way a user would be presented with them:
```
xoen, line 37: This may be a stupid question but why is it important to order "by flow"? (if it actually is)

I get that when comparing the current config vs. what's in NOMIS you want to normalise the two (and that sorts by code also for this reason) but when updating the NOMIS config does it matter?

I'm asking because:
1. in case I'm missing something
2. that sounds brittle (but it may be the way the API works)
3. if order is not actually important [for the purpose of updating the NOMIS configuration] you could re-use the [`Graph` class](https://github.com/ministryofjustice/hmpps-incident-reporting/blob/main/server/data/incidentTypeConfiguration/directedGraphs.ts) and its `dfs()` method which is already there as-is (but the results return the visited `Set`, so order is not retained)

server/data/incidentTypeConfiguration/nomisPayload.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+/** Body for `PUT /api/incidents/configuration/{nomisCode}` (updates an existing NOMIS type) */
+export function buildUpdateRequest(
+  dpsConfig: IncidentTypeConfiguration,
+  { description, active }: Pick<NomisSyncMeta, 'description' | 'active'>,
+): UpdateIncidentTypeConfigurationRequest {
+  return {
+    incidentTypeDescription: description,
+    active,
```
xoen, line 133: Would this suffice now that types are left active even when they "expire" (so that reports can be edited)?

I think this is the simplest thing we can do at the moment - and maybe the assumption is that incident types deactivation in NOMIS is done via a different route/mechanism?

The alternative is to also consider if the incident type is expired to determine whether it should be active in NOMIS.
That would be annoying because, when a new version comes online it would mean we'd have to sync to NOMIS before and after the rollover for both the old and the new version.

...having said that, given DPS is the source of true and no editing happen in NOMIS, does the active on the incident type even matter?

server/routes/admin/syncNomis/index.ts
```diff
@@ -0,0 +1,106 @@
...
+import { NotFound } from 'http-errors'
+
+import { buildCreateRequest, buildUpdateRequest } from '../../../data/incidentTypeConfiguration/nomisPayload'
+import { compareConfigs } from '../../../data/incidentTypeConfiguration/nomisCompare'
+import { types, isTypeActive } from '../../../reportConfiguration/constants'
+import { getIncidentTypeConfiguration } from '../../../reportConfiguration/types'
+import { errorResponseStatusMatches } from '../../../utils/utils'
+
+/** Active incident types offered for syncing, in display order */
+function activeTypes() {
```
xoen: I would personally prefer if the function signatures included the return type, I know it's optional in TypeScript, but IMHO it helps both readability and to catch some potential bugs.

(same goes for the `findActiveType()` function)

server/routes/admin/syncNomis/index.ts
```diff
@@ -0,0 +1,106 @@
...
+  res.render('pages/admin/syncNomis/confirm', {
+    type: typeMeta,
+    action: exists ? 'update' : 'create',
+    questionCount: request.questions.length,
+    prisonerRoleCount: request.prisonerRoles.length,
+  })
+}
+
+async function performSync(req: Request, res: Response): Promise<void> {
+  const typeMeta = findActiveType(req.params.dpsCode)
```
xoen: possibly `typeInfo` is a better name. Also, any reason why this is re-packaged into `meta`? Just to remove the `familyCode` property?

Given TypeScript is using structural typing, wouldn't `buildCreateRequest()`/`buildUpdateRequest()` just ignore the extra properties?

In other words, something like this should still work no?

```TypeScript
const typeInfo = findActiveType(req.params.dpsCode)
// ...
   const request = buildUpdateRequest(dpsConfig, typeInfo)
```

server/routes/admin/syncNomis/index.ts
```diff
@@ -0,0 +1,106 @@
...
+      const stored = await prisonApi.createIncidentTypeConfiguration(request)
+      res.render('pages/admin/syncNomis/result', {
+        type: typeMeta,
+        action: 'created',
+        comparison: compareConfigs(request, stored),
+      })
+    }
+  } catch (error) {
+    // A missing system-client role surfaces as 401/403 from Prison API
+    if (errorResponseStatusMatches(error, 403) || errorResponseStatusMatches(error, 401)) {
```
xoen: Do we need to also handle the `401 Unauthorized`? Wouldn't the API just respond `403 Forbidden` if the client is missing the right role?

A `401` would be an authentication error rather than an authorisation error 🤔 (e.g. wrong/expired token rather than missing role)

server/routes/admin/syncNomis/index.ts
```diff
@@ -0,0 +1,106 @@
...
+        action: exists ? 'update' : 'create',
+        roleError: true,
+      })
+      return
+    }
+    throw error
+  }
+}
+
+export default function makeSyncNomisRouter(): Router {
```
xoen, line 109: BTW, I like the way you structured this module/router.

server/views/pages/admin/syncNomis/result.njk
```diff
@@ -0,0 +1,56 @@
...
+{% from "govuk/components/panel/macro.njk" import govukPanel %}
+
+{% set pageTitle = "Sync result" %}
+
+{% block content %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      {% if roleError %}
```
xoen: Given this error message specifically mentions the client missing the role (and as the `roleError` variable name suggests), I'd argue we should _only_ set this to `true` if the API status code was a `403 Forbidden`.

Showing this in response to a `401 Unauthorized` would be misleading and possibly send us on the wrong path while investigating the error.

server/views/pages/admin/syncNomis/result.njk
```diff
@@ -0,0 +1,56 @@
...
+        {% if comparison.inSync %}
+          {{ govukNotificationBanner({
+            type: "success",
+            titleText: "Success",
+            html: "<p class=\"govuk-body\">NOMIS now matches the configuration held in this service.</p>"
+          }) }}
+        {% else %}
+          {{ govukNotificationBanner({
+            titleText: "NOMIS does not fully match this service",
+            html: "<p class=\"govuk-body\">The write succeeded, but the stored configuration differs from what was sent:</p>"
```
xoen: Just out of curiosity, what could result in this state? Just unfortunate timing with NOMIS manual changes or something like that?

server/data/incidentTypeConfiguration/nomisPayload.ts
Mjwillis: Its doesn't matter if one is left active in NOMIS as it's not used

server/views/pages/admin/syncNomis/result.njk
Mjwillis: I wanted to check what came back was what we sent as a precausion

server/data/incidentTypeConfiguration/nomisPayload.ts
Mjwillis: Think you misread it; the !answer.active check is traversal-only, nothing is filtered from the payload. And filtering would be bad: PUT does clear() + addAll under orphanRemoval = true, so dropping an inactive question deletes it and its answers from NOMIS and orphans historic reports.

server/data/incidentTypeConfiguration/nomisPayload.ts
Mjwillis: The order is the API contract, not decoration: mapQuestions assigns QUE_SEQ/LIST_SEQ from array index and reads sort by LIST_SEQ.

I think the graph parenthetical is wrong, too: JS Sets are insertion-ordered, so visited would have retained order

Mjwillis: The DfsResult.visited doesn't document an ordering contract,

xoen: Ah OK, that's why I was asking, it wasn't clear to me that the order was important.

Regarding JS's `Set` order you are right, given this is a property of JS I wouldn't bother  document it.

In theory then you could get rid of `orderedCodes` and just use `visited`:

```TypeScript
return [...visited, ...remaining].map([...])
```

also reads nicely.

You could also re-use the `Graph` class, I only played with this briefly but tests passed so maybe it's worth it? E.g., you could have something like this:

```TypeScript
export function orderQuestionsByFlow(dpsConfig: IncidentTypeConfiguration): QuestionConfiguration[] {
  const buildGraph = (): Graph<string> => {
    const graph = new Graph<string>()
    const questions = Object.values(dpsConfig.questions)
    for (const question of questions) {
      const activeAnswers = question.answers.filter(ans => ans.active === true)
      for (const answer of activeAnswers) {
        if (answer.nextQuestionCode) {
          graph.addEdge(question.code, answer.nextQuestionCode)
        }
      }
    }

    return graph
  }

  const graph = buildGraph()
  const { visited } = graph.dfs(dpsConfig.startingQuestionCode)
  // [...]
```

However is not exactly as the existing logic as it's **not** adding an edge for the last question as it doesn't have a subsequent question.
Alternatively you could do something similar to what done in other places and have the `Graph` contain `string | null` (and remove the `if (answer.nextQuestionCode) { ... }`, but then you'd have filter those `null` values out which is a bit of a chore.

Just an idea really, basically potentially the existing graph traversal logic could be reused.

server/data/incidentTypeConfiguration/nomisPayload.ts
xoen: > the !answer.active check is traversal-only

Gotcha! So we want to include all questions and answers - **_including inactive_** 👍

xoen (approved):

PR #908, opened 2026-07-22
Switch to @ministryofjustice/hmpps-azure-telemetry

jonesj2193:
Switch to using @ministryofjustice/hmpps-azure-telemetry instead of Application Insights for Azure telemetry as per the typescript template.


package.json
```diff
@@ -113,13 +113,13 @@
     "typescript": "^6.0.3",
     "yaml": "^2.9.0"
   },
-  "overrides": {
+"overrides": {
```
xoen, line 116: what happened to the indentation in this file? 🤔

xoen (approved):
FWIW LGTM

PR #912, opened 2026-07-23
IR-1765 Add Key or lock compromise v3 incident type and questionnaire

Mjwillis:
## What

Introduces `KEY_OR_LOCK_3`, the redesigned Key or lock compromise flow from the Safety team, switching over from v2 on 1 August 2026. This is the UI half of IR-1765; the type registration and migration are in the API repo (hmpps-incident-reporting-api#587), which should merge first.

A single triage question routes to a type/method pair for keys, locks, doors or gates, handcuffs, or any other locking device, and every branch finishes on a shared risk-level question — replacing the legacy NOMIS questionnaire's single mixed checkbox list along with its remedial-action and cost questions.

```
Q1 What type of key lock incident occurred?
 ├ Key          → type of key       → how compromised ─┐
 ├ Lock         → type of lock      → how compromised ─┤
 ├ Door or gate → type of door/gate → how compromised ─┼→ risk level → END
 ├ Handcuffs    → type of handcuffs → how compromised ─┤
 └ Other        → free text         → free text ───────┘
```

- New `KEY_OR_LOCK_3` config (questions, branching, mandatory "Provide details" comments) plus generated graph/diagram
- Registers the type and description, and a dated activation window retiring v2 as v3 begins (`activeTo`/`activeFrom` of 2026-08-01, matching the Tool loss v2 switch-over)
- Class of key/lock definitions supplied as answer hint text on the Class 1, Class 2 and Class 3 options of both type questions
- Prisoner involvement roles carried over from v2 unchanged

Every question is single-answer: the triage question branches, and a multiple-answer question whose answers lead to different next questions is rejected by `validateConfig()`. Question and answer codes are freshly allocated in the 9xxxxx block so the config can be back-synced to NOMIS without PK clashes.

Adds an optional `commentMultiline` flag to answer configuration so the two free-text questions render as textareas rather than single-line inputs, as the design requires. It defaults to the existing input everywhere else, with tests covering both paths.

## Copy source

Implemented from the approved Figma flow diagram. Where it differed from the earlier Jira table this resolved two wordings: `Lost or missing` (not `Lost/missing`) and `Removed or missing`, both consistent with the GOV.UK no-slashes rule applied in IR-1787.

## Testing

- `./scripts/validateConfig.ts` — no errors for `KEY_OR_LOCK_3`
- `./scripts/regenerateIncidentTypeGraphs.ts` — `.dot`/`.svg` regenerated; diff limited to the new files
- `npm run lint`, `npm run typecheck`, `npm test` (2747 tests) — green
- Rendered all seven pages through the real Nunjucks/GOV.UK stack: branching, page grouping (each type+method pair shares a page, so three screens per journey), the class hint text on the four class options per type question, the "Provide details" conditionals, and the two textareas all render as intended

## For design / follow-up

1. **Q3a has overlapping options** — `Attempted key snatch` and `Keys snatched (or attempted to be snatched)`. Implemented both as specified.
2. **"Hint text" placeholders** under several question headings in Figma read as unfilled — no `questionHint` set.


server/reportConfiguration/types/KEY_OR_LOCK_3.ts
```diff
@@ -0,0 +1,817 @@
...
+    '900009': {
+      code: '900009',
+      active: true,
+      question: 'key_type',
+      label: 'What type of key was compromised?',
+      multipleAnswers: false,
+      answers: [
+        {
+          code: '950047',
+          response: 'Brent keys',
```
xoen, line 96: Maybe a stupid question, and maybe everyone answering these questions would already know but...

What is a "brent key"?

xoen: If the answer is not obvious, it may be worth mentioning to DW/Content and add a _**little**_ bit of hint text to provide guidance for this answer.

server/reportConfiguration/types/KEY_OR_LOCK_3.ts
```diff
@@ -0,0 +1,817 @@
...
+        {
+          code: '950089',
+          response: 'Damaged',
+          active: true,
+          label: 'Damaged',
+          commentRequested: false,
+          commentMandatory: false,
+          dateMandatory: false,
+          nextQuestionCode: '900019',
+        },
```
xoen, line 580: Interesting how 'Damaged' (or potentially equivalent) is sometimes in the middle, sometimes as the first answer.

I wonder if these 'Damaged'-like options should be first for consistency as it may help with the user experience? (And maybe other damage-related - e.g. malicious/warn out - 2nd?)

It's almost like 'Other'/'Miscellaneous' in reverse.

server/reportConfiguration/types/KEY_OR_LOCK_3.ts
```diff
@@ -0,0 +1,817 @@
...
+          label: 'Level 5: No risk of disclosure. No relock.',
+          commentRequested: false,
+          commentMandatory: false,
+          dateMandatory: false,
+          nextQuestionCode: null,
+        },
+      ],
+    },
+  },
+  prisonerRoles: [
```
xoen, line 788: I wonder if something like 'Present at scene' would also make sense for this incident type.

xoen (approved):
Nice work Mike!

PR #913, opened 2026-07-24
IR-1765 Key or lock v3: consistent damage ordering and split key-snatch options

Mjwillis:
Follow-up to the merged Key or lock compromise v3 work (#912), applying business-lead (Ben) review feedback. Content-only edits to the `KEY_OR_LOCK_3` config — no question/answer codes change, so the NOMIS mapping is untouched. The type is not yet live (go-live 1 Aug 2026), so there is no data impact.

## Changes

**1. Lead each method question with its damage option**
The `Damaged` option previously sat in a different position per question (top in one, mid-list in others). The damage-type option now leads every method question for consistency:

| Method question | Option lifted to top |
|---|---|
| `key_compromise_method` | `Damaged` |
| `lock_compromise_method` | `Malicious damage` |
| `door_gate_compromise_method` | `Damaged` |
| `handcuffs_compromise_method` | `Damaged` (already first) |

**2. Split the overlapping key-snatch options**
`key_compromise_method` previously offered both `Attempted key snatch` and `Keys snatched (or attempted to be snatched)`, which overlapped. Now split cleanly into `Attempted key snatch` and `Key(s) snatched` (answer code unchanged).

## Testing

- `./scripts/validateConfig.ts` — "No errors found" for KEY_OR_LOCK_3
- Regenerated `KEY_OR_LOCK_3.dot` / `.svg`; only the three KEY_OR_LOCK_3 files change
- `npm run lint`, `npm run typecheck`, and the full Jest suite (2747 tests) pass


jonesj2193 (approved):
LGTM

PR #914, opened 2026-07-28
Vulnerabilities: js-yaml and @opentelemetry/core

jonesj2193:
To address major vulnerabilities: Added override relating @opentelemetry/core and updated js-yaml in package-lock. 

There is currently 1 major vulnerability left relating to brace-expansion but the patched version is only 5 days old, so need to wait until it has been released 7 days before we can use it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -524,22 +524,6 @@
         "node": ">=8.0.0"
       }
     },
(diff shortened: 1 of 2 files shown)


eve-bramley (approved):
LGTM

PR #919, opened 2026-07-31
IR-1812: Date filters bug on dashboard

jonesj2193:
There was an issue with the logic in the `processDateRange` function that meant no dates were parsed into the API call if only one of "date from" or "date to" had a value submitted in them. This led to no date filtering instead of setting a limit at one end. 

This PR updates the logic in the `processDateRange` so that when parsing the date filters without a valid value being set for the other date, it will still be be parsed and used in API call correctly.


cugorji-justice (dismissed):
LGTM

cugorji-justice (approved):
LGTM

